Issue #102, opened 2015-02-11
N+1 problem in CdoSnapshotFinder


pawelszymczyk:
Probably we should use one query instead of n+1 queries for download CdoSnapshots for given globalId.  

Here is link to loop that generates a problem:

https://github.com/javers/javers/blob/%2367_sql_repository/javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/CdoSnapshotFinder.java#L72

bartoszwalacik:
problem n+1 zapytań został zlikwidowany

PR #115, opened 2015-02-23
#102 n+1

bartoszwalacik:
many SQL optimizations,
few changes in javers-core, to minimize number of Repository interactions during commit()


bartoszwalacik:
przy okazji wchodzi

PR #145, opened 2015-05-04
Use Repository instead of CrudRepository

sebastian-sch:
In our case CrudRepository isn't not working because we directly extend the Repository. 

I think it is in general a good idea, to make it as flexible as possible.

[Review on Reviewable](https://reviewable.io/reviews/javers/javers/145)

### DIFF
--- a/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
+++ b/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
@@ -10,7 +10,7 @@ import org.javers.spring.auditable.AspectUtil;
 import org.javers.spring.auditable.AuthorProvider;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
-import org.springframework.data.repository.CrudRepository;
+import org.springframework.data.repository.Repository;
 import org.springframework.data.repository.core.RepositoryMetadata;
 import org.springframework.data.repository.core.support.DefaultRepositoryMetadata;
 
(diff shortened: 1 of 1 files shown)


bartoszwalacik:
but save() and delete() methods are defined in CrudRepository, Repository is only a marker interface

so is this pointcut typesafe?

```
@After("execution(public * save(..)) && this(org.springframework.data.repository.Repository)")
```

bartoszwalacik:
if you do not extend CrudRepository, how you define your save() and delete() methods?

bartoszwalacik:
@sebastian-sch are you there? :)

sebastian-sch:
Sorry for the delay, I was quite busy. The reason why I use the Repository instead of CrudRepository is, that I don't allow delete. I created my own CrudRepository with the same save() but without delete, for this the logic from JaversAuditableRepositoryAspect isn't working. 

Another option instead of changing    
```
@After("execution(public * save(..)) && this(org.springframework.data.repository.Repository)")
```
would be, to make JaversAuditableRepositoryAspect extendable  
to overwrite 
```
   @After("execution(public * delete(..)) && this(org.springframework.data.repository.Repository)")
    public void onDeleteExecuted(JoinPoint pjp)  {
        onVersionEvent(pjp, deleteHandler);
    }
```
or
```
    @After("execution(public * save(..)) && this(org.springframework.data.repository.Repository)")
    public void onSaveExecuted(JoinPoint pjp) {
        onVersionEvent(pjp, saveHandler);
    }
```

pawelszymczyk:
Hi Sebastian

Firstly thansk for a response, now your case is very clear. Aspectj does not allowed to extend concrete aspects, JaversAuditableRepositoryAspect is concrete so there is no possibility to change advices. Proper solution is to separate logic and advices, logic should be moved to abstract aspect, and advices should stay at contrete class as a default implementation, then you can extend abstract aspect and provide your own advices. Do you want to provide this solution in another pull request?

bartoszwalacik:
closing due to lack of activity

PR #151, opened 2015-05-12
Use startswith() to determine if method is getter

lalpert:
The function `isGetter()` checks if a method name begins with "get" or
"is". Previously, the code used `.substring(0,3)`, which throws an error
if the method name is fewer than 3 characters. `startswith()` won't
throw an error in that case and will instead return false.


pawelszymczyk:
Looks good to me but someone else must approve, @bartoszwalacik, @bgalek go on

bartoszwalacik:
that's better, thanks for the contribution

PR #158, opened 2015-05-22
first steps with hibernate proxied objects

bgalek:
hibernate test is passing, still need to fix javers-core to be able to run javers without declaring proxy (should be optional) and need to refactor default proxymanager impl, but you can see my aproach already

[Review on Reviewable](https://reviewable.io/reviews/javers/javers/158)


javers-core/src/test/groovy/org/javers/core/JaversCommitE2ETest.groovy
```diff
@@ -237,6 +237,23 @@ class JaversCommitE2ETest extends Specification {
         !secondCommit.diff.changes
     }
 
+    def "should unproxy"() {
```
bartoszwalacik: czy ten test coś testuje?

PR #160, opened 2015-05-25
Fixed deserialization of objects of a generic type with generic arguments

Dieler:
…arguments

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/type/ParametrizedDehydratedType.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/type/ParametrizedDehydratedType.java
@@ -19,7 +19,11 @@ class ParametrizedDehydratedType implements ParameterizedType{
         this.actualTypeArguments =
                 Lists.transform(javersType.getActualTypeArguments(), new Function<Type, Type>() {
                     public Type apply(Type input) {
-                        return typeMapper.getJaversType(input).getRawDehydratedType();
+                        JaversType javersType = typeMapper.getJaversType(input);
+                        if (!javersType.isGenericType()){
+                            return javersType.getRawDehydratedType();
+                        }
+                        return new ParametrizedDehydratedType(javersType, typeMapper);
                     }
                 }).toArray(new Type[]{});
     }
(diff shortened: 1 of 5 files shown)


javers-core/src/test/groovy/org/javers/core/metamodel/type/MapTypeTest.groovy
```diff
@@ -63,4 +66,17 @@ class MapTypeTest extends Specification{
         mType.keyType == String
         mType.valueType == Integer
     }
+
+    def "should return key & value Class if baseJavaType and value type are generic"(){
```
bartoszwalacik: this new test is passing also on master so we are missing something here...

Dieler: Ok, MapTypeTest is successful but TypeMapperTest fails without the fix in ParametrizedDehydratedType. Anyway, I've modified SnapshotRepositry and JaversRepositoryE2ETest to highlight the actual problem.

bartoszwalacik: thanks, you are right, `mapper.getDehydratedType()` should be "recursive"

javers-core/src/main/java/org/javers/core/metamodel/type/ParametrizedDehydratedType.java
```diff
@@ -19,7 +19,11 @@
         this.actualTypeArguments =
                 Lists.transform(javersType.getActualTypeArguments(), new Function<Type, Type>() {
                     public Type apply(Type input) {
-                        return typeMapper.getJaversType(input).getRawDehydratedType();
+                        JaversType javersType = typeMapper.getJaversType(input);
+                        if (!javersType.isGenericType()){
+                            return javersType.getRawDehydratedType();
+                        }
+                        return new ParametrizedDehydratedType(javersType, typeMapper);
```
bartoszwalacik: All I need to do is to manage the code repetition (with `typeMapper.getDehydratedType()`)

PR #188, opened 2015-08-12
Diff should implement Serializable otherwise I am not able to include…

klearchos:
… it inside a DTO object that already implements Serializable.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/diff/Diff.java
+++ b/javers-core/src/main/java/org/javers/core/diff/Diff.java
@@ -26,6 +26,7 @@ import static org.javers.common.validation.Validate.argumentIsNotNull;
  * @author bartosz walacik
  */
 public class Diff implements Serializable {
+    private static final long serialVersionUID = -7618674739116433849L;
     private final List<Change> changes;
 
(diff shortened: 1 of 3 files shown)


javers-core/src/main/java/org/javers/core/diff/Diff.java
```diff
@@ -24,7 +25,7 @@
  *
  * @author bartosz walacik
  */
-public class Diff {
+public class Diff implements Serializable {
```
bartoszwalacik: thanks for contributing to JaVers,
what about Diff components? Change, Atomic, etc. Think we need a simple test which proofs that Diff serialization works.
You could add it to JaversDiffE2ETest

xeraa:
Any update on this? I'm also having issues since Diff isn't serializable

bartoszwalacik:
@xeraa, as you see, @klearchos have created this Pull Request but he have abandoned the work ... 
you could continue this PR if you want, it's really easy.
Just add `serializable` for few classes and write an integration test

pawelszymczyk:
I polished it a bit in https://github.com/javers/javers/pull/269

Issue #199, opened 2015-09-17
javers compare is throwing exception when comparing two String arrays


dmmiller612:
Javers javers = JaversBuilder.javers().build();
        javers.compare(new String[]{"a", "b"}, new String[]{"a", "b"});

Currently the above code is throwing this exception:
 CLASS_NOT_MANAGED JaVers runtime error - given javaClass '[Ljava.lang.String;' is mapped to ArrayType, ManagedType expected .

I am currently using version 1.3.4.

dmmiller612:
Looks like this may be by design, which is completely fine since you can make a direct equal comparison. I just wanted to make sure.

bgalek:
I think you should compare entities (managed classes) here, exactly as exception says :)

bartoszwalacik:
Will check it , pretty  sure that works for lists

bartoszwalacik:
okay, so we have a feature for wrapping top level containers into Entities. It works for collections but not yet for Arrays.

If you would like to contribute a Pull Request, this issue is easy to solve in LiveGraphFactory.wrapTopLevelContainer()

dmmiller612:
sure I'll check it out.

bartoszwalacik:
released in v 1.3.8

PR #200, opened 2015-09-17
Issue #199 added Object array to LiveGraphFactory to stop throwing ex…

dmmiller612:
There are some design choices I made that could easily be changed. One for example, if you look at convertToObjectArray, I added the private function to handle primitive and Object arrays, which you may only want Object arrays.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/graph/LiveGraphFactory.java
+++ b/javers-core/src/main/java/org/javers/core/graph/LiveGraphFactory.java
@@ -46,8 +46,7 @@ public class LiveGraphFactory {
             return new SetWrapper((Set)handle);
         }
         if (handle.getClass().isArray()){
-            //return new ArrayWrapper(convertToObjectArray(handle));
-            return new ArrayWrapper((Object[])handle);
+            return new ArrayWrapper(convertToObjectArray(handle));
         }
         return handle;
     }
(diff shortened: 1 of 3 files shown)


bartoszwalacik:
thnx, will take a look at it tomorrow

bartoszwalacik:
could you give mie write access to https://github.com/dmmiller612/javers? I'd like to put some small changes in this PR

dmmiller612:
You should have write access, now.

javers-core/src/main/java/org/javers/core/graph/LiveGraphFactory.java
```diff
@@ -44,6 +45,10 @@ private Object wrapTopLevelContainer(Object handle){
         if (handle instanceof  Set){
             return new SetWrapper((Set)handle);
         }
+        if (handle.getClass().isArray()){
+            //return new ArrayWrapper(convertToObjectArray(handle));
+            return new ArrayWrapper((Object[])handle);
```
bartoszwalacik: @dmmiller612, I've removed this conversion and the tests are still passing, not sure what is it for ...
If this version is ok for you, I can merge

bartoszwalacik: @dmmiller612, I've removed this conversion and the tests are still passing, not sure what is it for ...
If this version is ok for you, I can merge

dmmiller612:
The rason I added the conversion is because you can't directly cast Object[] to a primitive[] at least in Java. I don't know much about groovy, and if it boxes the integer to an object. If you are fine just using objects, which is a valid design choice, I am okay with it.

bartoszwalacik:
you are right, seems copying is needed as there is no obvious way to do casting. So I've restored it.
I've added more test cases to cover both primitive and object arrays.

PR #244, opened 2015-11-02
For WildcardType, use Types of upper bounds (i.e. erase wildcard)

dbevacqua:
If I try to use JaVers to compare classes with upper-bounded wildcard types, I get an exception with message like the following:

```
beaconJaversException: GENERIC_TYPE_NOT_PARAMETRIZED JaVers runtime error - 
expected actual Class argument in type 'java.util.Set<? extends Foo>'. 
JaVers is strongly-typed and needs to know actual Class of elements stored in your collections. 
```

with an instruction to:

```
Try at least <Object>. Wildcards (e.g. <?>), unbounded type parameters (e.g. <T>) and raw types (e.g. List) are not supported.
```

For reasons, we cannot easily alter our model to narrow to `Set<Foo>` or `Set<SpecificSubclassOfFoo>`, and anyway we would prefer not to let our choice of this (admittedly wonderful) tool drive such things.

The error actually does not say that bounded parameters are not supported, so this pull request - treating wildcards with an upper bound as having the type of the upper bound - seems consistent with it. Certainly the tests still pass, and the behaviour is as we would expect.

In addition (and perhaps this should be the subject of another pull request), a more general change to include unbounded types (which seems like it's the same as "Try at least <Object>") might be worth considering, though such a change would break your tests (but only the ones that expect the GENERIC_TYPE_NOT_PARAMETRIZED error).

We've only recently started using JaVers, and our usage may be fairly narrow, so it's quite possible we've missed some obvious reason why this is not a good idea, but would appreciate if you would consider it.

### DIFF
--- a/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
+++ b/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
@@ -139,7 +139,19 @@ public class ReflectionUtil {
 
             if (t instanceof Class || t instanceof ParameterizedType) {
                 result.add(t);
+            } else if (t instanceof WildcardType) {
+                // If the wildcard type has an explicit upper bound (i.e. not Object), we use that
+                WildcardType wildcardType = (WildcardType) t;
+                if (wildcardType.getLowerBounds().length == 0) {
+                    for (Type type : wildcardType.getUpperBounds()) {
+                        if (type instanceof Class && ((Class<?>) type).equals(Object.class)) {
+                            continue;
+                        }
+                        result.add(type);
+                    }
+                }
             }
+
         }
 
         return Collections.unmodifiableList(result);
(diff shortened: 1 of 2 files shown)


bartoszwalacik:
You are right, upper-bounded wildcard should be treated as bound type, so your Pull goes in right direction. Thanks for contributing. Could you please add test case (in `CollectionTypeTest`) which will cover this feature?
like `"should treat wildcards with an upper bound as the type of its upper bound"()`

Considering your second issue about allowing unbounded wildcards, we could disqus it, but as you said, there should be separate issue/pull for this.

dbevacqua:
Great, thanks. I've added a test to this effect.

bartoszwalacik:
Excellent!

bartoszwalacik:
going to release it tomorrow morning

dbevacqua:
Great - thank you!

Issue #133, opened 2015-03-19
Restore orginal domain object from persisted snapshot


eladh:
Currently there is no possiblity to retrive original object or converte it from CDO.
In order to produce some kind of POJO based version management Eco-system we must be able to retrieve the original POJO.

Generally speaking i would love to see Javers  with version management API like :
1. Retrieve the Original POJO
2. Merge Operation => with Updated/Conflict status 
3. Restore Object to X version 

do you Store all Object graph or just Incremental information / diff from the last commit ?

10x

akarmazin:
1) For retrieving the original POJO as a quick and dirty solution you could use the class below (or something similar). 
see loadLatestEntityState() or loadEntityStateFromSnapshot() or loadEntityStateForCommitId()
However, as for now this solution has some restrictions (see class level comments).
2) For analyzing the conflict status you has to remember the commitId (and/or @Version value if exists) on POJO load operation in order to retrieve your base entity. By commiting the changed entity to your database you have to analyze the changes you maid against the base entity and changes already done by another user (actual entity state): here you could use Javers.compare for deep comparison, however the solution in case of deeply located elements in object graph would be not trivial
3) Restoring the whole object graph in relational database is not a trivial operation as well and mostly (or ever) is highly dependent on your business logic ->  So I don't think Javers could provide any general solution for this operation.

```
package javers.repository.test.reader;

import java.lang.reflect.Field;
import java.lang.reflect.Method;
import java.util.Collection;
import java.util.HashMap;
import java.util.HashSet;
import java.util.List;
import java.util.Map;
import java.util.Set;

import org.javers.common.collections.Optional;
import org.javers.common.reflection.JaversField;
import org.javers.core.Javers;
import org.javers.core.commit.CommitId;
import org.javers.core.metamodel.object.CdoSnapshot;
import org.javers.core.metamodel.object.InstanceId;
import org.javers.core.metamodel.object.InstanceIdDTO;
import org.javers.core.metamodel.property.Property;

/**
 * Loads under certain assumptions the original entity state for given
 * CdoSnapshot or CommitId.<br>
 * 1) Loads the whole object graph eagerly (lazy loading would require the use
 * of proxy objects). <br>
 * 2) All entities have constructor with no arguments -> Java Beans standard. <br>
 * 3) Each entity field ({@link JaversField}) has a setter method with the
 * argument of the same type as field -> Java Beans standard. <br>
 * 4) Since Javers lacks the function for retrieving the CdoSnapshot based on
[...]

PR #271, opened 2015-12-16
JaversSnapshotsCompiler added to core functionality

smirnp:
For more details see https://github.com/javers/javers/issues/133

### DIFF
--- a/javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
+++ b/javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
@@ -1,0 +1,217 @@
+package org.javers;
+
+import java.lang.reflect.Field;
+import java.lang.reflect.Method;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+import org.javers.common.collections.Optional;
+import org.javers.common.reflection.JaversField;
+import org.javers.core.Javers;
+import org.javers.core.commit.CommitId;
+import org.javers.core.metamodel.object.CdoSnapshot;
+import org.javers.core.metamodel.object.InstanceId;
+import org.javers.repository.jql.InstanceIdDTO;
+import org.javers.core.metamodel.property.Property;
+import org.javers.repository.jql.QueryBuilder;
+
+/**
+ * Loads under certain assumptions the original entity state for given
+ * CdoSnapshot or CommitId.<br>
+ * 1) Loads the whole object graph eagerly (lazy loading would require the use
+ * of proxy objects). <br>
+ * 2) All entities have constructor with no arguments -> Java Beans standard. <br>
+ * 3) Each entity field ({@link JaversField}) has a setter method with the
+ * argument of the same type as field -> Java Beans standard. <br>
+ * 4) Since Javers lacks the function for retrieving the CdoSnapshot based on
+ * nearest commitId -> all CdoSnapshots have to be loaded (assumes that limit =
+ * -1 means all snapshots, see implementation of
+ * {@linkplain #compileEntityStateForCommitId(InstanceId, String)}). <br>
+ * 5) Maps and other types are not yet supported. <br>
+ */
+public class JaversSnapshotsCompiler {
+
+    private static final double EPS = 0.001;
+    private Javers javers;
+
+    private Map<Object, Object> idToEntityMap = new HashMap<>();
+
+    public JaversSnapshotsCompiler(Javers javers) {
+        this.javers = javers;
+    }
+
+    public void clearCache() {
+        idToEntityMap.clear();
+    }
+
+    /**
+     * Loads original entity state for given CdoSnapshot
+     *
+     * @param snapshot
+     * @return
+     */
+    public Object compileEntityStateFromSnapshot(CdoSnapshot snapshot) {
+        clearCache();
+        try {
+            return compileEntityInternal(snapshot);
+        } catch (IllegalAccessException | InstantiationException e) {
+            e.printStackTrace();
+            return null;
+        }
+    }
+
+    /**
+     * Loads original entity state for given instanceId and commitId
+     *
+     * @param instanceId
+     * @param commitId
+     * @return
+     */
+    public Object compileEntityStateForCommitId(InstanceId instanceId,
+                                                String commitId) {
+        clearCache();
+        try {
+            return compileEntityForCommitId(instanceId, CommitId.valueOf(commitId));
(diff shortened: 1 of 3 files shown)


javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+        //List<CdoSnapshot> snapshots = javers.getStateHistory(convertInstanceIdToDTO(instanceId), -1); // load all
+        //List<CdoSnapshot> snapshots = javers.findSnapshots(QueryBuilder.byInstanceId(instanceId, instanceId.getgetBaseJavaClass() ).build());
+        List<CdoSnapshot> snapshots = javers.findSnapshots(QueryBuilder.byInstanceId(instanceId, instanceId.getClass()).build());
+
+        double commitToBeSearched = commitIdAsDouble(commitId);
+        CdoSnapshot snapFound = null;
+        // since all snapshot are sorted from the highest commit value to the
+        // lowest one
+        for (CdoSnapshot snap : snapshots) {
+            if (commitIdAsDouble(snap.getCommitId()) < (commitToBeSearched + EPS)) {
```
bartoszwalacik: commitId has well defined, decimal value, why to play with doubles and what is EPS?

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+    private Object compileEntityForCommitId(InstanceId instanceId, CommitId commitId) throws IllegalAccessException,
+            InstantiationException {
+
+        // TODO here the Javers function for loading the CdoSnapshot for the
+        // given commitId should be called, but the current Javers
+        // implementation doesn't provide it
+        // TODO assumes that limit = -1 means "all CdoSnapshots"
+        //List<CdoSnapshot> snapshots = javers.getStateHistory(convertInstanceIdToDTO(instanceId), -1); // load all
+        //List<CdoSnapshot> snapshots = javers.findSnapshots(QueryBuilder.byInstanceId(instanceId, instanceId.getgetBaseJavaClass() ).build());
+        List<CdoSnapshot> snapshots = javers.findSnapshots(QueryBuilder.byInstanceId(instanceId, instanceId.getClass()).build());
```
bartoszwalacik: fetching all snapshots could kill performance, we can think about adding queryByCommitId to JQL...

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+    /**
+     * Loads the latest commited entity state for given instanceId
+     *
+     * @param instanceId
+     * @return
+     */
+    public Object compileLatestEntityState(InstanceId instanceId) {
+        InstanceIdDTO dto = convertInstanceIdToDTO(instanceId);
+        Optional<CdoSnapshot> latestSnapshot = javers.getLatestSnapshot(dto, dto.getClass());
+        //Optional<CdoSnapshot> latestSnapshot = javers.getLatestSnapshot()getLatestSnapshot(convertInstanceIdToDTO(instanceId));
```
bartoszwalacik: all commented code should be removed

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+     * @param commitId
+     * @return
+     */
+    public Object compileEntityStateForCommitId(InstanceId instanceId,
+                                                String commitId) {
+        clearCache();
+        try {
+            return compileEntityForCommitId(instanceId, CommitId.valueOf(commitId));
+        } catch (IllegalAccessException | InstantiationException e) {
+            e.printStackTrace();
```
bartoszwalacik: public method in javers never returns null, just throw JaversRuntimeException

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+                        }
+                        for (Object element : toBeDeleted) {
+                            collection.remove(element);
+                        }
+                    } else if (propertyValue instanceof InstanceId) {
+                        propertyValueToSet = compileEntityForCommitId(
+                                (InstanceId) propertyValue, snap.getCommitId());
+                    }
+                    setter.invoke(instance, propertyValueToSet);
+                } catch (Exception exc) {
```
bartoszwalacik: swallowing all Exceptions is generally an antipatten, wrap it in JaversRuntimeException and rethrow

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
...
+import org.javers.repository.jql.InstanceIdDTO;
+import org.javers.repository.jql.QueryBuilder;
+
+/**
+ * Loads under certain assumptions the original entity state for given
+ * CdoSnapshot or CommitId.<br>
+ * 1) Loads the whole object graph eagerly (lazy loading would require the use
+ * of proxy objects). <br>
+ * 2) All entities have constructor with no arguments -> Java Beans standard. <br>
+ * 3) Each entity field ({@link JaversField}) has a setter method with the
```
bartoszwalacik: we should support both FIELD and BEAN property styles

javers-core/src/main/java/org/javers/snapshots-compiler/JaversSnapshotsCompiler.java
```diff
@@ -0,0 +1,216 @@
+package org.javers;
```
bartoszwalacik: there is no tests here, see how we write tests in Javers and add some
https://github.com/javers/javers/blob/master/javers-core/src/test/groovy/org/javers/core/JaversDiffE2ETest.groovy

bartoszwalacik:
@smirnp i did a code review, please follow my suggestions and then i could merge your solution to javers-incubating module

HarishBalachander:
Any timeline on when this feature will be released ?

smirnp:
I faced a problem, than on actual codes the return of "snapshots compiler" is not fully correct. So decicided to use not the latest codes. And I down't know, them will have an ability to modify it up to the actual codes.

bartoszwalacik:
of course, before we can merge, you need to catch up with our master

bartoszwalacik:
we have decided to create SnapshotsCompiler feature in a different way

PR #279, opened 2015-12-23
Small consistency fixes

mwesolowski:
Besides getComponents every other getter and adder in AbstractJavaBuilder class checks if container is already created.


pawelszymczyk:
Hi

Thanks for contribution :+1:

If you want to be linked as Javers contributor on our site: http://javers.org/  change commit author/e-mail, now change is commited by user unrecognized in github.

PR #280, opened 2015-12-24
JaversRepository's query methods accept QueryParams

mwesolowski:
Now we can add new parameters that will be passed to
query executors.


bartoszwalacik:
I will do CR this evening

mwesolowski:
This PR is now part of PR "Time range filtering".

PR #281, opened 2015-12-25
Fix of commit date storing in SQL DBs

mwesolowski:
LocalDateTime was converted to Timestamp with local timezone


mwesolowski:
I will create new pull request with this fix with test based on FakeDateProvider. 
This version is quite sloppy

PR #294, opened 2016-01-09
Add local version field in snapshots

milanov:
I'm happy to accept any comments, suggestions, etc :)

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
@@ -26,7 +26,7 @@ public final class CdoSnapshot extends Cdo {
     private final CdoSnapshotState state;
     private final SnapshotType type;
     private final List<String> changed;
-    private final Long version;
+    private final long version;
 
     /**
      * should be assembled by {@link CdoSnapshotBuilder}
@@ -37,7 +37,7 @@ public final class CdoSnapshot extends Cdo {
                 SnapshotType type,
                 List<String> changed,
                 ManagedType managedType,
-                Long version) {
+                long version) {
         super(globalId, managedType);
         Validate.argumentsAreNotNull(state, commitMetadata, type, managedType);
         this.state = state;
@@ -125,7 +125,7 @@ public final class CdoSnapshot extends Cdo {
         return type;
     }
 
-    public Long getVersion() {
+    public long getVersion() {
         return version;
     }
 }
--- a/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
@@ -125,6 +125,12 @@ public final class CdoSnapshot extends Cdo {
         return type;
     }
 
+    /**
+     * Object version number.<br/>
+     * Initial snapshot of given object has version == 1, next has version == 2.
+     *
+     * @since 1.4.3
+     */
     public long getVersion() {
         return version;
     }
--- a/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
@@ -26,6 +26,7 @@ public final class CdoSnapshot extends Cdo {
     private final CdoSnapshotState state;
     private final SnapshotType type;
     private final List<String> changed;
+    private final Long version;
 
     /**
      * should be assembled by {@link CdoSnapshotBuilder}
@@ -35,13 +36,15 @@ public final class CdoSnapshot extends Cdo {
                 CdoSnapshotState state,
                 SnapshotType type,
                 List<String> changed,
-                ManagedType managedType) {
+                ManagedType managedType,
+                Long version) {
         super(globalId, managedType);
         Validate.argumentsAreNotNull(state, commitMetadata, type, managedType);
         this.state = state;
         this.commitMetadata = commitMetadata;
         this.type = type;
         this.changed = changed;
+        this.version = version;
     }
 
     /**
@@ -120,5 +123,9 @@ public final class CdoSnapshot extends Cdo {
 
     public SnapshotType getType() {
         return type;
+    }
+
+    public Long getVersion() {
+        return version;
     }
 }
--- a/javers-core/src/test/groovy/org/javers/core/json/typeadapter/CdoSnapshotTypeAdapterTest.groovy
+++ b/javers-core/src/test/groovy/org/javers/core/json/typeadapter/CdoSnapshotTypeAdapterTest.groovy
@@ -5,9 +5,9 @@ import groovy.json.JsonSlurper
 import org.javers.core.commit.CommitId
 import org.javers.core.commit.CommitMetadata
 import org.javers.core.metamodel.object.CdoSnapshot
-import org.javers.repository.jql.ValueObjectIdDTO
 import org.javers.core.model.DummyUser
 import org.javers.core.model.DummyUserDetails
+import org.javers.repository.jql.ValueObjectIdDTO
 import org.javers.test.builder.DummyUserDetailsBuilder
 import org.joda.time.LocalDateTime
 import spock.lang.Specification
@@ -162,7 +162,7 @@ class CdoSnapshotTypeAdapterTest extends Specification {
         snapshot.version == 5L
     }
 
-    def "should deserialize CdoSnapshot.version to 0 when version field is missing"() {
+    def "should deserialize CdoSnapshot.version to 1 when version field is missing"() {
 
         given:
         def json = new JsonBuilder()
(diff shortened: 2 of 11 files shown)


javers-core/src/main/java/org/javers/core/metamodel/object/CdoSnapshot.java
```diff
@@ -26,6 +26,7 @@
     private final CdoSnapshotState state;
     private final SnapshotType type;
     private final List<String> changed;
+    private final Long version;
```
bartoszwalacik: Think we dont want version to be null, so maybe primitive  long?

milanov: Ok, I'll change that. Do you prefer for me to simply commit this change or to squash it with the previous (I didn't find contribution guidelines)?

bartoszwalacik: True, we need to write contribution guidelines.  Just push another commit but wait a while,  i will take a deeper look at this PR at the evening

javers-core/src/test/groovy/org/javers/core/json/typeadapter/CdoSnapshotTypeAdapterTest.groovy
```diff
@@ -41,6 +40,7 @@ class CdoSnapshotTypeAdapterTest extends Specification {
         json.globalId.entity == "org.javers.core.model.DummyUser"
         json.globalId.cdoId == "kaz"
         json.type == "INITIAL"
+        json.version == 1
```
bartoszwalacik: I suggest to test it in a different (minimal) way.
So test new version field only in 2 basic tests:
`def "should serialize CdoSnapshot to Json"()` and
`def "should deserialize CdoSnapshot"()`
and not to change other tests.
Deserializer should not break if field is missing (and it will be missing if users ugrade to new JaVers version and will load snapshots created in old JaVers version).

javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
```diff
@@ -39,6 +39,7 @@
     public static final String SNAPSHOT_COMMIT_FK =    "commit_fk";
     public static final String SNAPSHOT_GLOBAL_ID_FK = "global_id_fk";
     public static final String SNAPSHOT_TYPE =         "type";
+    public static final String VERSION =               "version";
```
bartoszwalacik: unfortunately, there is no db schema update in PolyJDBC (the tool we use to manage sql),
we do it manually, see ex `JaversSchemaManager.alterCommitIdColumnIfNeeded()`

So we need a new method which checks if column not exists and then runs DDL.
It's not easy to write and test.
There is another way we can try. Version field can be moved to CdoSnapshotState, which is serialized to JSON in all repositories. If so, no need to touch repositories code (only core JsonTypeAdapter).

milanov: Okay, I went this way but I'm a bit stuck on how to serialize the CdoSnapshotState including the new version field. Currently it is serialized as `state: {prop1: val1, ..}`. I can either add the version field inside of the state as another property (but then it could already have a property with the same name which means a conflict) or serialize it as `state: {version: X, properties:  {prop1: val1, ..}}`, but this wouldn't be backward compatible. Any ideas on this?

bartoszwalacik: You are right, there is no elegant way to serialize version in CdoSnapshotState. Looks like we need to come back to your original solution and write `addColumnIfNotExists()` method in `JaversSchemaManager`. We need to take a deep dive into PollyJDBC code. It's very lightweight but simple library ...

bartoszwalacik: There are two possible solutions. We can write generic addColumnIfNotExists() method. Better place for this is PollyJDBC, but easier way is to write it in JaVers.
Second solution, would be quick hack, just a series of Ifs with proper DDLs,
like we did in line 57 in JaversSchemaManager

milanov: There will be a problem with the previously stored snapshots, because they would have to all be traversed and assigned versions in incrementing order..This might not be that easy.

bartoszwalacik: True, updatin old snapshots would be very problematic

bartoszwalacik: But we dont necessary need to do full migration.  version ia a ne feature, so user would have sequence : [null, null, 1, 2] :(

milanov: Hi, I updated the PR the way we discussed it, but I'm still not sure if we can't update the previous records in the jv_snapshots with the corresponding versions. What is the field that is the same for all snapshots of a given entity (global_id_fk)? I'll try some combination of join/update/count query :)

bartoszwalacik: ok, but maybe we leave this migration script as an option? We can describe it in release notes.
For large databases this kind of update could be really slow. For sure we need to add support for nulls (nulls can be treated as 0).

For users interested in new field, migration script (run manually) could update 
version sequences : [null, null, 1, 2] -> [1,2,3,4]

What do you think about this solution?

milanov: I don't know whether it will be slow, since I don't have that much experience with really large databases, but your proposal sounds good and reasonable :)
Two more questions - do you have any other comments to be fixed and the one I asked in the previous comment - which is the field that is the same for all snapshots of a given entity.

bartoszwalacik: I dont have more commments,
besides the old one to CdoSnapshotTypeAdapterTest.groovy which is not applied yet?

Considering your second question *which is the field that is the same for all snapshots of a given entity.*
There are no special columns in SQL tables which are not reflected in Java (besides primary and foreign keys).

If you are thinking about SQL for data migration script,
snapshots can be sorted asc by primary key and updated: `SET version = rownum` 
But rownum is dialect-specific. In Oracle - `rownum`, in Postgres -  `row_number()` and so on ...

I think that troubles with data migration script should not stop us from releasing `getVersion()`
as soon as we prepare clear javadoc 
(that will describe the issue with nulls in old snapshots versions to JaVers users)

milanov: Here's a query that could serve as a migration script:
`UPDATE jv_snapshot s SET s.version = (SELECT COUNT(*) + 1 FROM (SELECT * FROM jv_snapshot) s2 WHERE s.global_id_fk = s2.global_id_fk and s2.snapshot_pk < s.snapshot_pk)`
It is SQL-vendor agnostic and accomplishes the task. I'm not that sure about performance, but that could be described in the release notes.

bartoszwalacik: I think that one sub-select could be removed:

```
UPDATE jv_snapshot s SET s.version = (
SELECT COUNT(*) + 1 FROM jv_snapshot s2 WHERE s.global_id_fk = s2.global_id_fk and s2.snapshot_pk < s.snapshot_pk)
```

This query will be slow, but maybe good enough for migration script.

So PR is now ready to merge?

milanov: Now it should be, I separated the test according what we discussed. As a note to the query from your last comment - I also tried it at first, but it gives an error in MySQL ("You can't specify target table 's' for update in FROM clause", as well as in PostgreSQL("ERROR: column "s" of relation "employees" does not exist") and I haven't tested it on the others.

bartoszwalacik: ok, give us a few more days for merge and release

bartoszwalacik: @milanov there are few bugs here:
- NPE is thrown when version field is missing in a snapshot stored in SQL or Mongo (created by javers <= 1.4.2)
migration script is optional, so JaVers cant crash when someone doesnt run it
(I will fix it)
- version is ALWAYS deserialized to 1
take a look at line 60 in CdoSnapshotTypeAdapter. When snapshots are deserialized from DB, previous field is null, 
and builder overwrites number set in `.withVersion(version)`
- version field is never write to Sql database, you simply forgot to add it to `CdoSnapshotRepository.insertSnapshot()`

bartoszwalacik: see https://github.com/javers/javers/commit/12a8afff29e2e4e0f19289515cd7fcac4049e7dc

bartoszwalacik:
I did fixes on my branch, and just merged all into master

bartoszwalacik:
released in 1.4.4
http://javers.org/documentation/features/#release-notes

Issue #309, opened 2016-01-18
JQL - new CommitId filter


bartoszwalacik:
Requested by @atomfrede
https://github.com/deepu105/generator-jhipster-entity-audit/issues/4

Query by CommitId could be implemended as a new filter, available only for snapshot queries:

```
javers.findSnapshots( QueryBuilder.byInstanceId("bob", Employee.class).withCommitId(1.0).build() )
```

atomfrede:
I can try to contribute the new filter if you like.

bartoszwalacik:
We have a team meeting today evening, and we will do some planning. Stay tuned

atomfrede:
Awesome! Thanks for your support!

bartoszwalacik:
@atomfrede, @mwesolowski is going to implement this new filter. Stay tuned

bartoszwalacik:
released in JaVers v 1.4.5
see http://javers.org/documentation/jql-examples/#commit-id-filter

bartoszwalacik:
@atomfrede check new filter by commitId(). This filter makes sense only for snapshot queries

PR #315, opened 2016-01-19
#309 Commit id filter

mwesolowski:


mwesolowski:
I have to check why this change fails in integration environment. There was no such problem with Fongo,

Issue #322, opened 2016-01-26
JQL - new withVersion() filter


bartoszwalacik:
new filter for snapshot queries:

`javers.findSnapshots( QueryBuilder.byInstanceId("bob", Employee.class).withVersion(5).build() )`

`withVersion()` should work simmilary to `withCommitId()` filter

requested by:
@atomfrede
@milanov

bartoszwalacik:
It could be implemented in the same way like `commitId` filter done by @mwesolowski
see https://github.com/javers/javers/pull/318/files

bartoszwalacik:
@milanov, looks like @mwesolowski already did it :)

atomfrede:
@bartoszwalacik @mwesolowski Wow, that was fast! Impressive!

bartoszwalacik:
released in v. 1.4.7
see http://javers.org/documentation/jql-examples/#version-filter

bartoszwalacik:
@atomfrede check out new `withVersion()` filter, does it solve your issue?

PR #323, opened 2016-01-26
#322 JQL - snapshot version filter

mwesolowski:


mwesolowski:
I don't know if it is not too late now but maybe we could think about wrapping snapshot version primitive into some object?

bartoszwalacik:
imho long for snapshot.version is fine

javers-core/src/main/java/org/javers/repository/api/QueryParamsBuilder.java
```diff
@@ -89,11 +90,19 @@ public QueryParamsBuilder commitId(CommitId commitId) {
         return this;
     }
 
+    /*
+     * limits results to Snapshots with a given version
+     */
+    public QueryParamsBuilder version(Long version) {
```
bartoszwalacik: imo we should use `long` not `Long` here, the same as in `CdoSnapshot.getVersion()`

bartoszwalacik: I will change it

bartoszwalacik: ok, int this context null means, dont use it in query

PR #329, opened 2016-01-30
328 add core module to compile dependencies

pawelszymczyk:


pawelszymczyk:
I will add it into another PR.

Issue #355, opened 2016-04-07
JaversAuditableRepositoryAspect does not properly handle thrown exceptions


Drizzt321:
Javers Spring support uses JaversAuditableRepositoryAspect on Spring-Data `@Repository` classes for `@JaversAuditable` or the Spring-Data `CrudRepository` interface. If an exception is thrown from one of those methods, then JaversAuditableRepositoryAspect, which uses `@After` will still execute and try and perform a versioning. According to [Spring AOP Docs](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/aop.html), `@After` will always execute, regardless of if there was an exception thrown or not. `@AfterReturning` instead executes only if the method returns normally.

Here's the [original StackOverflow question](https://stackoverflow.com/questions/36465724/spring-aop-javersauditable-still-trying-to-commit-a-change-when-exception-has) for reference.

bartoszwalacik:
fixes merged to master, waiting for release

bartoszwalacik:
released in 1.6.2

PR #356, opened 2016-04-07
Issue #355 JaversAuditableRepositoryAspect does not properly handle thrown exceptions

Drizzt321:
Solution is to use @AfterReturning which only runs on a normal method exit, and @AfterThrowing which checks if @Transactional was used on the @Repository method, and if noRollbackFor or noRollbackForClassName is configured (still commit even if these Exceptions/Throwables are thrown) and matches the current Throwable, still proceed with the versioning.

I looked at updating JaversAuditableRepositoryAspectTest, but frankly I'm unsure how I'd even go about testing this kind of behavior, especially given I'm completely unfamiliar with Spock and only somewhat know Groovy. If you need me to update the test I'll be happy to, but some direction on what documentation bits might be useful for me to start with would be greatly appreciated.

### DIFF
--- a/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
+++ b/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
@@ -2,9 +2,7 @@ package org.javers.spring.auditable.aspect;
 
 import org.aspectj.lang.JoinPoint;
 import org.aspectj.lang.annotation.AfterReturning;
-import org.aspectj.lang.annotation.AfterThrowing;
 import org.aspectj.lang.annotation.Aspect;
-import org.aspectj.lang.reflect.MethodSignature;
 import org.javers.common.collections.Optional;
 import org.javers.core.Javers;
 import org.javers.spring.annotation.JaversSpringDataAuditable;
@@ -13,20 +11,21 @@ import org.javers.spring.auditable.AuthorProvider;
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.data.repository.core.RepositoryMetadata;
 import org.springframework.data.repository.core.support.DefaultRepositoryMetadata;
-import org.springframework.transaction.annotation.Transactional;
-
-import java.lang.annotation.Annotation;
 
 /**
- * Creates three @After advices.
+ * Creates three @AfterReturning advices.
  * <br/><br/>
  *
- * Commits all arguments passed to methods with @JaversAuditable annotation.
+ * Commits all arguments passed to methods with @JaversAuditable annotation only if the method exits normally, e.g. no
+ * Throwable has been thrown. If the method is also annotated with @Transactional and the noRollbackFor or
+ * noRollbackForClassName are used, no commit will be done as the method still throws a Throwable. A manual
+ * {@link Javers#commit(String, Object)} can be used instead.
  * <br/><br/>
  *
  * For spring-data Repositories with @JaversSpringDataAuditable annotation: <br/>
  * - commits all arguments passed to save() methods,  <br/>
  * - commits delete of arguments passed to delete() methods.  <br/>
+ * - commits only when a method exits normally without a Throwable having been thrown.
  */
 @Aspect
 public class JaversAuditableRepositoryAspect {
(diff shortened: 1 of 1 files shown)


javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
```diff
@@ -93,4 +120,55 @@ private void applyVersionChange(RepositoryMetadata metadata, Object domainObject
...
+        if( transactionalAnnotation == null ) {
+            return false;
+        }
+
+        boolean executeCommit = false;
+        Class<? extends Throwable> exceptionClass = ex.getClass();
+        //grab the @Transactional noRollbackFor classes
+        Class<? extends Throwable>[] noRollbackForClasses = transactionalAnnotation.noRollbackFor();
+        if( noRollbackForClasses != null && noRollbackForClasses.length > 0 ) {
+            for( Class<? extends Throwable> aClass : noRollbackForClasses ) {
```
bartoszwalacik: This looks quite complex and I think it doesnt handle exceptions hierarchy.
In Spring you can declare `noRollbackFor=SomeException.class` and it won't rollback for all subclasses of SomeException.

I'm not sure if we really need this `@AfterThrowing` feature. Maybe just `@AfterReturning` and javers.commit() only when a method is executed without any exception?

Drizzt321:
So I didn't do exception hierarchy because I don't believe the rollbackFor does. I test since handily the exception I was seeing had a nested exception (Spring-data wrapper with the MySql exception inside) and the @Transactional only committed for the top level exception. 

For subclasses, I didn't check how @Transactional worked, so I can go update that after testing.

bartoszwalacik:
This is a general purpose aspect and it works also for Mongo and I dont like the idea of adding strong copuling to Spring JPA Transaction Manager here.
Why not just change `@After` to `@AfterReturning`? Will it solve your issue?
If we do so, JaVers logic will be simple and predictible, no javers.commit() when an adviced method thrown an exception.

Drizzt321:
Mongo methods can't have Spring's @Transactional on it? Or looking at it another way, if there's no @Transactional, the extra checks are not used. And I'm not using JPA either, I'm using MyBatis which ends up with just a DataSource Connection from Spring, but it's still using Spring's Transaction management features.

I'm fine removing the @AfterThrowing, it just seemed that it could apply to other folks who may be using that noRollbackFor. I don't believe anything we'll be doing will need that, but if people do use it and Javers at the same time, then they might not get the results the expect.

bartoszwalacik:
I can accept this @AfterThrowing feature but only without duplicating Spring code. I mean if you can access current Spring transaction and check if it is rollbackOnly, thats fine. Otherwise please remove @AfterThrowing, lets keep it simple.

Drizzt321:
Not sure what you mean "only without duplicating Spring code". I'm not adding any additional library dependencies if that is what you mean.

bartoszwalacik:
No, I mean copying Spring logic concering handling noRollbackFor annotation parameter.
As I said before, try to check rollbackOnly flag from the current Spring transaction or remove the whole afterThrowing thing.

Drizzt321:
Hm, yes, as I read more about it I see your point with regards to the transaction TransactionAspectSupport.currentTransactionStatus.isRollbackOnly(). However, in the testing I just did we can't rely on that because the @AfterThrowing happens before the Spring Transaction has a chance to check receive the exception and setRollbackOnly(). 

So either there's some kind of fancy transaction listener/caching the entity to version until it commits, which I don't think would be a good idea, or either the @AfterThrowing is left as is or completely removed. Unless there's another option I haven't thought of.

bartoszwalacik:
ok, so remove it please

Drizzt321:
Ok, removed, added some Javadocs to document the changes.

javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversAuditableRepositoryAspect.java
```diff
@@ -1,29 +1,31 @@
...
 import org.springframework.data.repository.core.support.DefaultRepositoryMetadata;
 
 /**
- * Creates three @After advices.
+ * Creates three @AfterReturning advices.
  * <br/><br/>
  *
- * Commits all arguments passed to methods with @JaversAuditable annotation.
+ * Commits all arguments passed to methods with @JaversAuditable annotation only if the method exits normally, e.g. no
+ * Throwable has been thrown. If the method is also annotated with @Transactional and the noRollbackFor or
```
bartoszwalacik: obsolete javadoc

bartoszwalacik:
@Drizzt321 , your commits are merged to master in another PR - https://github.com/javers/javers/pull/360 (rebased), so this PR can be closed

Issue #375, opened 2016-05-27
Support for commit properties in auto-audit aspect


bartoszwalacik:
Users want to pass commit properties when using auto-audit aspect.
It can be implemented in the similar way to `AuthorProvider` bean

bartoszwalacik:
released in 2.0.0-RC2

PR #376, opened 2016-05-27
https://github.com/javers/javers/issues/375

bartoszwalacik:
Support for commit properties in auto-audit aspect


mwesolowski:
+1

Issue #380, opened 2016-06-09
2.0.0-RC1 does not work with complex objects


pranasblk:
```
Result: 13:23:09.398 [main] INFO  o.j.j.Java8AddOns - loading Java8 add-ons ...
13:23:09.420 [main] INFO  o.j.g.GroovyAddOns - loading Groovy add-ons ...
13:23:09.425 [main] INFO  o.j.c.m.s.ScannerModule - using FIELD mappingStyle
13:23:09.484 [main] INFO  o.j.c.JaversBuilder - using fake InMemoryRepository, register actual implementation via JaversBuilder.registerJaversRepository()
13:23:09.506 [main] INFO  o.j.c.JaversBuilder - JaVers instance is up & ready
JaversException: CLASS_EXTRACTION_ERROR JaVers bootstrap error - Don't know how to extract Class from type 'T[]'.
	at org.javers.common.reflection.ReflectionUtil.extractClass(ReflectionUtil.java:207)
	at org.javers.core.metamodel.type.TypeMapperState.findNearestAncestor(TypeMapperState.java:167)
	at org.javers.core.metamodel.type.TypeMapperState.infer(TypeMapperState.java:160)
	at org.javers.core.metamodel.type.TypeMapperState.access$000(TypeMapperState.java:23)
	at org.javers.core.metamodel.type.TypeMapperState$1.apply(TypeMapperState.java:87)
	at org.javers.core.metamodel.type.TypeMapperState$1.apply(TypeMapperState.java:85)
	at org.javers.core.metamodel.type.TypeMapperState.computeIfAbsent(TypeMapperState.java:118)
	at org.javers.core.metamodel.type.TypeMapperState.getJaversType(TypeMapperState.java:85)
	at org.javers.core.metamodel.type.TypeMapper.getJaversType(TypeMapper.java:101)
	at org.javers.core.metamodel.type.TypeMapper.getPropertyType(TypeMapper.java:165)
	at org.javers.core.graph.ObjectGraphBuilder$1.apply(ObjectGraphBuilder.java:108)
	at org.javers.core.graph.ObjectGraphBuilder$1.apply(ObjectGraphBuilder.java:106)
	at org.javers.core.metamodel.type.ManagedClass.getManagedProperties(ManagedClass.java:60)
	at org.javers.core.metamodel.type.ManagedType.getProperties(ManagedType.java:49)
[...]

PR #386, opened 2016-06-10
#380 Reproduce issue using Java 8 and Arrays of Generics.

pranasblk:


bartoszwalacik:
ok, thnx for this repro, we are working on the bug fix here https://github.com/javers/javers/pull/387

PR #387, opened 2016-06-11
#380 generic arrays

mwesolowski:


pranasblk:
Bardzo Dobzo
Dzenkue!

PR #408, opened 2016-07-25
Change Object Equality

DanielJonesEB:
I've added some more object equality implementations, as I was having difficulty testing that my usage of Javers was creating the expected changes. 

I hope the Spock tests are acceptable - not written any before, so I was just trying to copy the style that was elsewhere.


bartoszwalacik:
CR fixes https://github.com/javers/javers/pull/409/commits/b5ad741e4811510c071e6707072c908e1dd48918

PR #415, opened 2016-08-11
Fix TypeMapperState.mappedTypes. 

hank-cp:
Class loaded by different classLoader are not equal, so using Class as map key will have mare than one entry in that map. Use type name instead.

The failed test case could be reproduce when work with Spring-Boot 1.3+ and DevTools.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapperState.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapperState.java
@@ -21,7 +21,7 @@ import static org.javers.common.validation.Validate.argumentsAreNotNull;
  * @author bartosz.walacik
  */
 class TypeMapperState {
-    private final Map<Type, JaversType> mappedTypes = new ConcurrentHashMap<>();
+    private final Map<String, JaversType> mappedTypes = new ConcurrentHashMap<>();
     private final Map<DuckType, Class> mappedTypeNames = new ConcurrentHashMap<>();
     private final TypeFactory typeFactory;
     private final ValueType OBJECT_TYPE = new ValueType(Object.class);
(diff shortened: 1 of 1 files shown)


bartoszwalacik:
thanks for the PR, does current impl (class as a map key) causes problems or this is only about optimization (limiting the map size) ?

hank-cp:
@bartoszwalacik yes, class as map key causing problem. Class loaded by different classloader is not the same instance.

bartoszwalacik:
Ok but what kind of problem does it caus in javers?

hank-cp:
I capture these screenshots from debugger.

image

image

My Javers is build like this:
`
    JaversUtil.getCommonJaversBuilder()
                    .registerValueTypeAdapter(new NGGeneralDictionaryTypeAdapter()))
                    .build();
`

bartoszwalacik:
@hank-cp I understand the issue, you have two classes loaded by two classloaders. These classes have the same name but might have different properties.
I dont understand why it is not OK for you. Why you want to map two (possibly different) classes to the same JaVers type?

hank-cp:
@bartoszwalacik I don't mean to add the second JaversType (with EntityType). I guess it's added automatically when Javers, when it's trying to find the best matching JaveryType for my class. In this case, Javers couldn't found my custom ValueTypeAdapter by the key, which is loaded by another classloader, so it wired EntityType for the best guess, thus cause my problem.

bartoszwalacik:
ok, got it

bartoszwalacik:
now I'm on holiday, I will merge this in next week

bartoszwalacik:
one fix 
https://github.com/javers/javers/pull/420/commits/1a9e2ea27f41d4d877c0624cbff955597547b6c4
(we don't want to break Java7 compatibility)

bartoszwalacik:
released in 2.1.2

PR #425, opened 2016-09-05
fixing ShallowReference

bartoszwalacik:


javers-core/src/test/groovy/org/javers/core/cases/OpBoxCaseClassExtractionError.groovy
```diff
@@ -0,0 +1,70 @@
...
+import org.javers.core.JaversBuilder
+import org.javers.core.metamodel.annotation.Id
+import org.javers.core.metamodel.annotation.ShallowReference
+import org.javers.core.metamodel.clazz.EntityDefinitionBuilder
+import spock.lang.Specification
+
+/**
+ * @author bartosz.walacik
+ */
+public class OpBoxCaseClassExtractionError extends Specification {
```
bgalek: OpboxCase Class? :)

bartoszwalacik: indeed :)

PR #426, opened 2016-09-06
Add aspect for JpaRepository#saveAndPublish

ouaibsky:
I simply add new point cut for JpaRepository.

I think with 
> @Target annotation 

some code is useless (#getRepositoryInterface). If you OK I can put @Target for other point cut and remove this code into another commit.

Christophe

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,4 +3,5 @@ fongoVersion=2.0.7
 springVersion=4.3.2.RELEASE
 springBootVersion=1.4.0.RELEASE
 hibernateVersion=5.0.9.Final
+aspectjweaverVersion=1.8.6
 spockVersion=1.0-groovy-2.4
--- a/javers-spring-boot-starter-sql/build.gradle
+++ b/javers-spring-boot-starter-sql/build.gradle
@@ -2,6 +2,8 @@ dependencies {
     compile project(':javers-core')
     compile project(':javers-spring')
     compile project(':javers-persistence-sql')
+
+    compile "org.aspectj:aspectjweaver:$aspectjweaverVersion"
     //should be on application's classpath when using JaversSpringBootStarterSql
     optional "org.springframework.boot:spring-boot-starter-data-jpa:$springBootVersion"
     optional "org.springframework.boot:spring-boot-configuration-processor:$springBootVersion"
--- a/javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSpringDataJpaAuditableRepositoryAspect.java
+++ b/javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSpringDataJpaAuditableRepositoryAspect.java
@@ -1,0 +1,34 @@
+package org.javers.spring.boot.sql;
+
+import org.aspectj.lang.JoinPoint;
+import org.aspectj.lang.annotation.AfterReturning;
+import org.aspectj.lang.annotation.Aspect;
+import org.javers.core.Javers;
+import org.javers.spring.auditable.AuthorProvider;
+import org.javers.spring.auditable.CommitPropertiesProvider;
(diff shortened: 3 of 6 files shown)


bartoszwalacik:
I dont get it. What is `saveAndPublish()`?
Why you **replacing** `JaversSpringDataAuditableRepositoryAspect` with your `JaversSpringDataJpaAuditableRepositoryAspect` ?

ouaibsky:
I didn't replace, I extend It because JpaRepository inherits CrudRepository:
[JpaRepository](https://github.com/spring-projects/spring-data-jpa/blob/master/src/main/java/org/springframework/data/jpa/repository/JpaRepository.java)

bartoszwalacik:
still, i dont get it. What is saveAndPublish()?

Firts, please try to describe what feature/use case you are trying to add to javers, then we will talk about implementation

bartoszwalacik:
why not just one new method, in existing class - `JaversSpringDataAuditableRepositoryAspect` ?
```
    @AfterReturning("execution(public * saveAndFlush(..)) && this(org.springframework.data.jpa.repository.JpaRepository)")
    public void onSaveAndFlushExecuted(JoinPoint pjp) {
        onVersionEvent(pjp, saveHandler);
    }
```

ouaibsky:
Because If we are doing that, in **spring-core** you'll have to put a dependency on **spring-boot-starter-data-jpa** (otherwise an exception is throwing when you instantiate the bean)

bartoszwalacik:
I've closed, because this code is plain wrong (misspelled method name) and over-complicated.
I expect sth simple, just like I've put in a snippet above

ouaibsky:
And you don't care about dependency ?

bartoszwalacik:
yes, we care, that's why we use `optional`. Did you tried this? And not dep to `spring-boot-starter-data-jpa` but `spring-data-jpa`

ouaibsky:
I know what is optional dep.
but it will no help us because if you add in JaversSpringDataAuditableRepositoryAspect:

    @AfterReturning("execution(public * saveAndFlush(..)) && this(org.springframework.data.jpa.repository.JpaRepository)")
    public void onSaveAndFlushExecuted(JoinPoint pjp) {
        onVersionEvent(pjp, saveHandler);
    }

Then as I tell you when you'll instantiate this bean, spring-aop will throw an exception because it will not find JpaRepository in your dependencies.

>      Caused by:
        org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'org.springframework.context.event.internalEventListenerProcessor': Initialization of bean failed; nested exception is java.lang.IllegalArgumentException: warning no match for this type name: org.springframework.data.jpa.repository.JpaRepository [Xlint:invalidAbsoluteTypeName]

PS: Don't run test in IJ because it's using dep of your whole project (child included)

**So you don't have the choice to put spring-data-jpa as a mandatory dep.**

I know my PR was not simple as expected but this it doesn't break current dependencies schema.

If you are ok to change dep mgt, of course my first try was to add this new advice in JaversSpringDataAuditableRepositoryAspect.

If you want something of very very simple, let's just put   @ConditionalOnMissingBean 

    @Bean
    @ConditionalOnMissingBean
    public JaversSpringDataAuditableRepositoryAspect javersSpringDataAuditableAspect(Javers javers, AuthorProvider authorProvider, CommitPropertiesProvider commitPropertiesProvider) {
        return ...
    }

Christophe.

bartoszwalacik:
ok, right, thats how it should be done:
- new aspect in `javers-spring` named `JaversSpringDataJpaAuditableRepositoryAspect` with `onSaveAndFlushExecuted(...)`
- common code (with `JaversSpringDataAuditableRepositoryAspect`) moved to abstrcat class
- optional & test dependency to `spring-data-jpa` in `javers-spring`
- new aspect delcared in spring-boot starter for JPA (`JaversSqlAutoConfiguration`)
- tests for new aspect
- doc update (that's what I can manage)

agreed?

Issue #263, opened 2015-12-05
@TypeName annotation scanner


bartoszwalacik:
As a JaVers user I want to scan all my classes with @TypeName annotation (in given package)
with one call to `JaversBuilder` (preferably `scanTypeNames(String packageToScan)`) 
in order to use them in all kinds of JQL queries without getting TYPE_NAME_NOT_FOUND exception.

*@Since 1.4*

Vote here if you think that it could be useful.

abibell:
**Symptom:** I still get the error.

I used [JaversSqlAutoConfiguration.java](https://github.com/javers/javers/blob/master/javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSqlAutoConfiguration.java) and added my `.scanTypeName(Employment.class).build();`. Upon debugging my `javers(ConnectionProvider connectionProvider)` doesn't get hit, however my other properties like `commitPropertiesProvider()` and `springSecurityAuthorProvider()` get hit.

Am I missing some configuration? I am using JaversSqlAutoConfiguration.java

```java
	
package com.test.config;
import java.util.HashMap;
import java.util.Map;

import javax.persistence.EntityManagerFactory;

import org.hibernate.SessionFactory;
import org.hibernate.dialect.Dialect;
import org.hibernate.engine.spi.SessionFactoryImplementor;
import org.javers.core.Javers;
import org.javers.core.MappingStyle;
import org.javers.core.diff.ListCompareAlgorithm;
import org.javers.hibernate.integration.HibernateUnproxyObjectAccessHook;
import org.javers.repository.sql.ConnectionProvider;
import org.javers.repository.sql.DialectName;
import org.javers.repository.sql.JaversSqlRepository;
import org.javers.repository.sql.SqlRepositoryBuilder;
import org.javers.spring.auditable.AuthorProvider;
import org.javers.spring.auditable.CommitPropertiesProvider;
import org.javers.spring.auditable.MockAuthorProvider;
import org.javers.spring.auditable.SpringSecurityAuthorProvider;
import org.javers.spring.auditable.aspect.JaversAuditableRepositoryAspect;
import org.javers.spring.boot.sql.DialectMapper;
import org.javers.spring.boot.sql.JaversProperties;
import org.javers.spring.jpa.JpaHibernateConnectionProvider;
import org.javers.spring.jpa.TransactionalJaversBuilder;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;
[...]

PR #430, opened 2016-09-07
@ConditionalOnMissingBean on javers(connectionProvider) for Mongo and Sql

abibell:
Fix for http://stackoverflow.com/questions/39322714/javerssqlautoconfiguration-javers-not-getting-called

### DIFF
--- a/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
+++ b/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
@@ -9,11 +9,15 @@ import org.javers.common.validation.Validate;
 import org.javers.core.Javers;
 import org.slf4j.Logger;
 
+import java.io.File;
+import java.io.IOException;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.*;
 import java.math.BigDecimal;
+import java.net.URL;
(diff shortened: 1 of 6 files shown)


bartoszwalacik:
@abibell I thought that this PR is about adding `@ConditionalOnMissingBean` to javers starters configuration. Then you could replace default javers bean with your javers bean with your configuration (`scanTypeName()`).
We didnt talked abould implementing `scanTypeNames()`. This is another issue - https://github.com/javers/javers/issues/263
And I think that javers should use some dedicated lib for classpath scanning.

bartoszwalacik:
The first commit in this PR is OK, the second is not OK. I will merge if you remove the second commit. I think, the first one is enough to solve your issue

abibell:
Oops, 2nd commit wasn't meant to come into PR. Try it now.

With regards to second issue, what would library do you have in mind to use, is it worth it? Do you see introducing the classpath scanning library can be a separate issue to getting scanTypeNames(package)?

bartoszwalacik:
yes, there is a separate issue for that - https://github.com/javers/javers/issues/263
Some research is needed, the ann scanning lib should be small (no dependencies) and fast

bartoszwalacik:
released in 2.2.2

Issue #398, opened 2016-07-05
Define schema for javers related tables


ianagius:
At the moment, JaVers is creating tables in the public schema. It would be great to be able to define the schema where these tables are created,

bartoszwalacik:
Agreed but schema support have to be implemented in polyJdbc first

ouaibsky:
It could be nice to provide some configuration as well in order to tune table name.
I'm working for a big company, where DBA's rules are strict and we must follow some guidelines on naming (mainly table)

Christophe

ianagius:
has there been any progress on this case? We just need to expose a setter method for private String schema property in SchemaInspectorImpl.

bartoszwalacik:
no progress yet. Dont think that someone from core team will take this issue soon. If you need this feature - contribute

bartoszwalacik:
see also https://github.com/polyjdbc/polyjdbc

ianagius:
We have decided to contribute :) and we have created a pull request for you to consider. Let us know if you have any questions or you would like further clarifications.

bartoszwalacik:
I can't accept this, see https://github.com/javers/javers/pull/433#pullrequestreview-435527
It will be easier to work on this issue, if you follow my guidelines. As I wrote in the second comment in this issue: schema support have to be implemented in polyJdbc first.

LucasDesenv:
This is a good one... It's not quite nice to use public schema for auditing.. kinda security failure, no? Or I didn't get it?

bartoszwalacik:
security failure? I'm affraid that you are missing that JaVers doesn't write to **public** schema. It writes to a default schema, for most databases it means schema of a current db user

LucasDesenv:
Oh, I get it! Thanks for explaining.

bartoszwalacik:
done & released in 2.4.0

pedroxs:
How to configure the schema? I could not find anywhere on the docs page

bartoszwalacik:
@pedroxs docs updated, see https://javers.org/documentation/repository-configuration/#sql-databases

pedroxs:
Perfect, thanks!

PR #433, opened 2016-09-16
#398 Define schema for javers related tables 

ianagius:
added withSchema() to SqlRepositoryBuilder to enable users to create and use tables in a defined Schema.

As part of the change, the static table and sequence names in the FixedSchemaFactory class have been made private and access to them is made available through static getters (which check if a schema was defined).

The JaversSchemaManager has been modified to check for the javers tables if a schema has been defined.  If no schema has been defined, this resposibility to handed over to PolyJDBC.

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/SqlRepositoryBuilder.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/SqlRepositoryBuilder.java
@@ -16,6 +16,9 @@ import java.sql.SQLException;
 public class SqlRepositoryBuilder extends AbstractContainerBuilder {
     private static final Logger logger = LoggerFactory.getLogger(SqlRepositoryBuilder.class);
 
+    public static String SCHEMA_NAME = "";
+    public static final String SCHEMA_TABLE_SEP = ".";
+
     private DialectName dialectName;
     private ConnectionProvider connectionProvider;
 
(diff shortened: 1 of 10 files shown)


ianagius:
With regards to 1 and 2, we totally agree and will amend accordingly.

Regarding 3, we really didn't want to get involved in another project.  Having said that, we are taking a look at polyjdbc code. 

Still unsure about your suggestion to make PolyJDBC aware of the schema and not make any changes in JaVers when we consider code in JaVers like the following (found in SnapshotFilter):
```
static final String COMMIT_WITH_SNAPSHOT = 
     SNAPSHOT_TABLE_NAME 
     + " INNER JOIN " 
     + COMMIT_TABLE_NAME 
     + " ON " + COMMIT_PK 
     + " = " + SNAPSHOT_COMMIT_FK;

static final String COMMIT_WITH_SNAPSHOT_GLOBAL_ID =
     COMMIT_WITH_SNAPSHOT 
     + " INNER JOIN " + GLOBAL_ID_TABLE_NAME + " g ON g." 
     + GLOBAL_ID_PK + " = " + SNAPSHOT_GLOBAL_ID_FK 
     + " LEFT OUTER JOIN " + GLOBAL_ID_TABLE_NAME 
     + " o ON o." + GLOBAL_ID_PK + " = g." + GLOBAL_ID_OWNER_ID_FK;
```

What we can't understand is how these queries will be handled by PolyJDBC.   These queries look like they are being passed to PolyJDBC which in turn executes them immediately.  Are you suggesting that PolyJDBC rewrites these queries as they're being passed by JaVers? If this is so, then PolyJDBC needs to be aware of the tables in JaVers so this rewriting can occur.  Or are we missing something?

bartoszwalacik:
ok, you are right, both sides (PolyJdbc and JaVers should be schema-aware) to some extent.
1. PolyJdbc's primary responsibility is schema creation (DDL's). If you pass a schema name to PolyJdbc instance, it could create all objects in this concrete schema.
2. I don't want any major changes in `JaversSchemaManager.java`, PolyJdbc should manage it
3. I'm also a committer of PolyJdbc, so I can help you to go through a Pull Request to PolyJdbc
4. You are right that JaVers can add schema prefix to table names (rewriting FROM clauses in PolyJdbc would be awkward)
5. the solution you proposed (`getGlobalIdTableName()`, etc.) can work but these methods can't be static (various JaVers instances can have various schema names).

hope we now agree on the design

ianagius:
yes definitely. will come back to you with a new pull request

ianagius:
created a pull request on polyjdbc side. Will create a pull request for JaVers once that is accepted.

bartoszwalacik:
ok, seems like a good plan

ianagius:
created an other pull request

bartoszwalacik:
thanks, I will take a look

PR #437, opened 2016-10-03
Update QueryBuilder::withNewObjectChanges(boolean) method to use parameter

gtiwari333:
It seems the parameter `withNewObjectChanges `was missed to pass to `queryParamsBuilder`.

```
    public QueryBuilder withNewObjectChanges(boolean newObjectChanges) {  // newObjectChanges not passed to queryParamsBuilder
        queryParamsBuilder.newObjectChanges(true);
        return this;
    }
```


bartoszwalacik:
thanks for the fix

PR #438, opened 2016-10-06
Added Schema support

ianagius:
Updated PolyJDBC library to have polyjdbc schema support
Added method in SqlRepositoryBuilder .withSchema() to be able to define a schema
Created TableNameManager to cater for table names which is used throughout the system.

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
@@ -1,5 +1,6 @@
 package org.javers.repository.sql.schema;
 
+import org.javers.repository.sql.pico.TableNameManager;
 import org.polyjdbc.core.dialect.Dialect;
 import org.polyjdbc.core.dialect.OracleDialect;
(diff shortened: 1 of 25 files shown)


javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
```diff
@@ -103,10 +109,10 @@ private Schema commitPropertiesTableSchema(Dialect dialect, String tableName) {
         Schema schema = new Schema(dialect);
         RelationBuilder relationBuilder = schema.addRelation(tableName);
         relationBuilder
-            .primaryKey(tableName + "_pk").using(COMMIT_PROPERTY_COMMIT_FK, COMMIT_PROPERTY_NAME).and()
+            .primaryKey(normaliseSchemaTableName(tableName) + "_pk").using(COMMIT_PROPERTY_COMMIT_FK, COMMIT_PROPERTY_NAME).and()
```
bartoszwalacik: we don't want to alter names of existing db objects

bartoszwalacik:
I did CR fixes on a new branch
https://github.com/javers/javers/commit/43318781a839d5b2c99b03d2511cb6a35929a2a5

Issue #435, opened 2016-09-22
Deactivating @JaversSpringDataAuditable by configuration


arnzel:
Hi.I would like to activate/deactivate the functionality of the Annotation @JaversSpringDataAuditable by a configuration property.  

Can the contructor of "JaversSpringDataAuditableRepositoryAspect" be changed so that i can commit an own implementation of "OnSaveAuditChangeHandler". So the method "handle" can only invoke the commit only  if the configuration property is active.

Or is there a better wy for my intention ?

bartoszwalacik:
You mean in spring boot starter?

arnzel:
Yes in the spring boot starter

bartoszwalacik:
Could you describe your use case? I wonder if this feature would be useful for other javers users

arnzel:
Hi. I got some MongoRepositories to which i added the Annotation "JaversSpringDataAuditable". The Annotation does what it should do, but for a reason, i dont know yet, our software is obviously slower with these Annotations. So i would like to deactivate  the creating the snapshots of the entities until i find the problem. At the moment i comment out these annotations. A better solution wouldbe that i can add a property in my spring boot config with the name "auditEnabled" and then add some logic which activates/deactivates the creating of snapshots depending on the value of this property.

Or in other words: it would be good to deactivate/activate the Annotations by only changing one line.

I hope the intention is clear now.

bartoszwalacik:
Right, think about a PR. It looks like a small change in spring boot starters configuration

bartoszwalacik:
see https://github.com/javers/javers/blob/master/javers-spring-boot-starter-mongo/src/main/java/org/javers/spring/boot/mongo/JaversMongoAutoConfiguration.java

bartoszwalacik:
fixed in release 2.6.0
see http://javers.org/release-notes/

PR #440, opened 2016-10-11
#435 Create JaversSpringDataAuditable with own OnSaveAuditChangeHandler

arnzel:


javers-spring/src/main/java/org/javers/spring/auditable/aspect/springdata/JaversSpringDataAuditableRepositoryAspect.java
```diff
@@ -27,6 +27,11 @@ public JaversSpringDataAuditableRepositoryAspect(Javers javers, AuthorProvider a
         this(new OnSaveAuditChangeHandler(javers, authorProvider, commitPropertiesProvider),
                 new OnDeleteAuditChangeHandler(javers, authorProvider, commitPropertiesProvider));
     }
+
```
bartoszwalacik: why another constructor?

arnzel:
As i described in #435 i would like to use an own instance of OnSaveAuditChangeHandler which saves changes only when a given configuration property is set

bartoszwalacik:
ok, but it should be implemented in our spring-boot starters `JaversSqlAutoConfiguration` and `JaversMongoAutoConfiguration` using `@ConditionalOnProperty`

arnzel:
okay you mean adding this annotation to "JaversSpringDataAuditableRepositoryAspect" ?

bartoszwalacik:
No, I mean adding a new spring property and disable aspect when this property is false

bartoszwalacik:
implemented by core team
see http://javers.org/release-notes/

PR #443, opened 2016-10-17
Avoid nested DiffIgnored ValueObject being compare

hank-cp:
Set gsonBuilder exclusionStrategies to avoid nested ValueObject circular reference.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
+++ b/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
@@ -3,6 +3,7 @@ package org.javers.core.json;
 import com.google.gson.*;
 import org.javers.common.validation.Validate;
 import org.javers.core.json.typeadapter.date.DateTypeCoreAdapters;
+import org.javers.core.metamodel.annotation.DiffIgnore;
 
 import java.lang.reflect.Type;
 import java.util.Collection;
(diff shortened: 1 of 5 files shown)


bartoszwalacik:
why this strategy? javers doesn't serialize user's domain objets

hank-cp:
Sorry, I didn't explain this issue clear enough. The problem is not about comparison, but about committing against MongoRepository. I add a few test case, hope that would explain the problem.

bartoszwalacik:
so, it's more gson problem than javers issue

bartoszwalacik:
you have a wrong mapping for CrossReferenceObjectA.
Obviously, this is a complext object so you should map it as ValueObject (not Value)
It's enough to remove @Value ann, see
https://github.com/javers/javers/commit/7ee9f6de22263f41be17a11a136e718ecd442415

hank-cp:
@bartoszwalacik Actually I do need to mark CrossReferenceObjectA/B as `@Value`, three reasons in my thought:
1. In my case, CrossReferenceObjectA/B are very general, fundamental class, most of other business class will refer them, and they change quite a lot and often.
2. These objects basically have a little bit complex, cross reference relation graph, according to my observation, committing them as `@ValueObject`, the javers repository size will be glowing very fast, a little bit out of control.
3. I have to implement CustomPropertyComparator for these classes so it could compare by my custom algorithm rather than Javers general comparison algorithm. I think this will required `@Value` or `JaversBuilder.javers().registerValue()` (correct my if I am wrong).

bartoszwalacik:
Ok, makes sense, if @Value fits better to your model then you should be able to choose it. Give me a while and I will merge

bartoszwalacik:
released in 2.5.0

PR #450, opened 2016-10-27
448 add new annotation called @IgnoreDeclaredProperties

ianagius:
This is our suggested solution to have @IgnoreAllProperties annotation that ignores all properties of the annotated class.
Tells us what you think.


bartoszwalacik:
released in 2.7.0

PR #451, opened 2016-10-27
 #435 auditable aspects deactivation

mwesolowski:


arnzel:
Thanks for the commit.

PR #452, opened 2016-10-28
448 Add @IgnoreDeclaredProperties annotation

ianagius:
Add @IgnoreDeclaredProperties to classes to ignore all their properties.
If a class is annotated with @IgnoreDeclaredProperties and is part of an audited hierarchy, only the properties of the class are ignored, and JaVers will still track any other properties in the class hierarchy.
By contrast, if a class is annotated with @DiffIgnore JaVers will ignore all instances of that class.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/scanner/FieldBasedPropertyScanner.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/scanner/FieldBasedPropertyScanner.java
@@ -2,6 +2,7 @@ package org.javers.core.metamodel.scanner;
 
 import org.javers.common.reflection.JaversField;
 import org.javers.common.reflection.ReflectionUtil;
+import org.javers.core.metamodel.annotation.IgnoreDeclaredProperties;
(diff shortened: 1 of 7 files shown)


javers-core/src/main/java/org/javers/core/metamodel/scanner/FieldBasedPropertyScanner.java
```diff
@@ -20,14 +21,22 @@ public FieldBasedPropertyScanner(AnnotationNamesProvider annotationNamesProvider
 
     @Override
     public PropertyScan scan(Class<?> managedClass) {
+        final IgnoreDeclaredProperties ignoreDeclaredPropertiesAnnotation = managedClass.getAnnotation(IgnoreDeclaredProperties.class);
```
bartoszwalacik: annotations scanning should be done in ClassScan. I'll fix it

bartoszwalacik:
moved this thread to https://github.com/javers/javers/pull/453

PR #453, opened 2016-11-05
Ianagius 448

bartoszwalacik:


javers-core/src/main/java/org/javers/core/metamodel/scanner/FieldBasedPropertyScanner.java
```diff
@@ -20,14 +21,22 @@ public FieldBasedPropertyScanner(AnnotationNamesProvider annotationNamesProvider
...
     public PropertyScan scan(Class<?> managedClass) {
+        final IgnoreDeclaredProperties ignoreDeclaredPropertiesAnnotation = managedClass.getAnnotation(IgnoreDeclaredProperties.class);
         List<JaversField> fields = ReflectionUtil.getAllPersistentFields(managedClass);
         List<Property> propertyList = new ArrayList<>(fields.size());
 
-        for (JaversField field : fields) {
-            boolean hasTransientAnn = field.hasAnyAnnotation(annotationNamesProvider.getTransientAliases());
-            boolean hasShallowReferenceAnn = field.hasAnyAnnotation(annotationNamesProvider.getShallowReferenceAliases());
-            propertyList.add(new Property(field, hasTransientAnn, hasShallowReferenceAnn));
+        if (ignoreDeclaredPropertiesAnnotation != null) {
```
bartoszwalacik: this is wrong, it means that when `@IgnoreDeclaredProperties` is present -- all properties will be ignored

javers-core/src/main/java/org/javers/core/metamodel/scanner/PropertyScanner.java
```diff
@@ -5,6 +5,14 @@
  *
  * @author pawel szymczyk
  */
-interface PropertyScanner {
-    PropertyScan scan(Class<?> managedClass);
+abstract class PropertyScanner {
+    abstract public PropertyScan scan(Class<?> managedClass, boolean ignoreDeclaredProperties);
+
+    public PropertyScan scan(Class<?> managedClass, ClassAnnotationsScan classScan) {
```
mwesolowski: I'm not sure if PropertyScanner should be aware of ClassAnnotationsScan. Is it not an unnecessary coupling? The only information needed here is whether declared properties should be ignored.

bartoszwalacik: ok, will fix it

PR #458, opened 2016-11-15
fix type erasure issue with Bean MappingStyle

ianagius:
The proposed solution is to ignore bridge methods in the method JaversMethodFactory.getAllMethods().
Two test models and 1 test to test this particular scenario has been added.

### DIFF
--- a/javers-core/src/test/java/org/javers/core/model/AbstractEntity.java
+++ b/javers-core/src/test/java/org/javers/core/model/AbstractEntity.java
@@ -1,0 +1,10 @@
+package org.javers.core.model;
+
+import java.io.Serializable;
+
+/**
+ * Created by Ian Agius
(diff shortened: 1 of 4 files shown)


ianagius:
Contribution for issue 457

javers-core/src/test/java/org/javers/core/model/AbstractEntity.java
```diff
@@ -0,0 +1,10 @@
+package org.javers.core.model;
+
+import java.io.Serializable;
+
+/**
+ * Created by Ian Agius
+ */
+public abstract class AbstractEntity<ID extends Serializable> {
```
bartoszwalacik: these classes could be embedded in a groovy test, I will fix it

Issue #461, opened 2016-11-16
Delete with no revisions present throws


gmadkat:
I am using JaVers in my Spring application, and when I am trying to delete an entity which has no revision history, I am seeing this exception: JaversException CANT_DELETE_OBJECT_NOT_FOUND: failed to delete object <my object>, it doesn't exists in JaversRepository
I looked at your code and noticed that is is intentional, do you have a workaround or would you consider allowing a way for us to delete entities without revisions? I am able to do a PUT (update) on an entity without revision history though, just not a delete.

Code I think this is happening in:
 ```
 public Commit createTerminalByGlobalId(String author, Map<String, String> properties, GlobalId removedId){
        Validate.argumentsAreNotNull(author, properties, removedId);
        Optional<CdoSnapshot> previousSnapshot = javersRepository.getLatest(removedId);
        if (previousSnapshot.isEmpty()){
            throw new JaversException(JaversExceptionCode.CANT_DELETE_OBJECT_NOT_FOUND,removedId.value());
        }
        CommitMetadata commitMetadata = nextCommit(author, properties);
        CdoSnapshot terminalSnapshot = snapshotFactory.createTerminal(removedId, previousSnapshot.get(), commitMetadata);
        Diff diff = diffFactory.singleTerminal(removedId, commitMetadata);
        return new Commit(commitMetadata, Lists.asList(terminalSnapshot), diff);
    }
```

We do need to support historical data with no revision history and delete them as needed so please help us do this. 

Thank you

**Comment by gmadkat** [comment]
I am happy to make any changes and submit a pull request if you are ok accepting them.

**Comment by bartoszwalacik** [comment]
And these deletes of non existing objects are called by aspects (@JaversSpringDataAuditable) ?

**Comment by gmadkat** [comment]
Yes. We have a Postgres database of legacy data with the JpaRepository objects annotated with @JaversSpringDataAuditable, and we were trying to delete them.

**Comment by bartoszwalacik** [comment]
Ok, in this case it makes sense to swallow and log the exception. We will accept a PR, please implement this catch in OnDeleteAuditChangeHandler (in aspects impl).

**Comment by gmadkat** [comment]
In order to fix,  I am trying to build JaVers using gradlew build on Mingw Windows and got an error:

```
org.javers.core.json.JsonConverterUtilTypesTest > should convert class java.io.File (\tmp\file.txt) to and from JSON FAILED
    Condition not satisfied:

    javers.jsonConverter.toJson( givenValue ) == expectedJson
[...]

PR #474, opened 2016-12-06
Changes to support adding a revision for legacy tables

gmadkat:
Changes to not throw for adding a revision where there is no prior history. We have legacy tables with no JaVers revisioning and need to support adding new history for them


bartoszwalacik:
This is wrong according to what we have agreed on https://github.com/javers/javers/issues/461 :
_So instead of swallowing the exception we will check if given object exists and call javers.commitShallowDelete() only if given object exists_

As a rule, JaVers API methods, never returns null.

I will implement it myself in `OnDeleteAuditChangeHandler`

PR #477, opened 2016-12-07
https://github.com/javers/javers/issues/461

bartoszwalacik:


javers-spring/src/main/java/org/javers/spring/auditable/aspect/springdata/OnDeleteAuditChangeHandler.java
```diff
@@ -18,14 +20,22 @@
...
+    public void handle(RepositoryMetadata repositoryMetadata, Object domainObjectOrId) {
         Map<String, String> props = commitPropertiesProvider.provide();
         String author = authorProvider.provide();
 
-        if (isIdClass(repositoryMetadata, domainObject)) {
-            javers.commitShallowDeleteById(author, instanceId(domainObject, repositoryMetadata.getDomainType()), props);
-        } else if (isDomainClass(repositoryMetadata, domainObject)) {
-            javers.commitShallowDelete(author, domainObject, props);
+        if (isIdClass(repositoryMetadata, domainObjectOrId)) {
+            if (javers.findSnapshots(QueryBuilder.byInstanceId(domainObjectOrId, repositoryMetadata.getDomainType()).build()).size() == 0) {
```
jefftaillon: Instead of using javers.findSnapshots() I would try to use a more efficient way of checking if snapshots exists. Maybe javers.getLatestSnapshot() is more appropriate as it will not load as much data !? Or maybe consider implementing some  exists() method.

bartoszwalacik: Yes, javers.getLatestSnapshot() would be better

Issue #475, opened 2016-12-07
org.polyjdbc.core.exception.SequenceLimitReachedException: [SEQUENCE_LIMIT_REACHED] Sequence jv_commit_pk_seq has reached its limit


achimbitzer:
We encountered this error while running multithreaded performance tests on our application, which uses Javers for versioning of database changes. See below for full stack trace.

I think this looks like an error inside or while accessing the polyjdbc class Sequence without proper synchronisation for thread safety, so this may be a bug inside polyjdbc. Therefore I've created an issue for polyjdbc, too: https://github.com/polyjdbc/polyjdbc/issues/22

Do you think this assumption for the cause of the exception is correct?

Suggestions for fixing this in Javers:
- either update polyjdbc when this issue is fixed in polyjdbc
- or implement a workaround to  ensure thread safety in this case.

`org.polyjdbc.core.exception.SequenceLimitReachedException: [SEQUENCE_LIMIT_REACHED] Sequence jv_commit_pk_seq has reached its limit of 8899. Before fetching value, check if recalculation is needed using recalculationNeeded() method.
	at org.polyjdbc.core.key.Sequence.nextValue(Sequence.java:50) ~[Sequence.class:?]
	at org.polyjdbc.core.key.SequenceAllocation.generateKey(SequenceAllocation.java:55) ~[SequenceAllocation.class:?]
	at org.polyjdbc.core.query.TransactionalQueryRunner.insert(TransactionalQueryRunner.java:108) ~[TransactionalQueryRunner.class:?]
	at org.javers.repository.sql.repositories.CommitMetadataRepository.insertCommit(CommitMetadataRepository.java:46) ~[CommitMetadataRepository.class:?]
	at org.javers.repository.sql.repositories.CommitMetadataRepository.save(CommitMetadataRepository.java:34) ~[CommitMetadataRepository.class:?]
	at org.javers.repository.sql.JaversSqlRepository.persist(JaversSqlRepository.java:63) ~[JaversSqlRepository.class:?]
	at org.javers.repository.api.JaversExtendedRepository.persist(JaversExtendedRepository.java:109) ~[JaversExtendedRepository.class:?]
	at org.javers.core.JaversCore.commit(JaversCore.java:80) ~[JaversCore.class:?]
	at org.javers.spring.jpa.JaversTransactionalDecorator.commit(JaversTransactionalDecorator.java:59) ~[JaversTransactionalDecorator.class:?]
[...]

PR #479, opened 2016-12-12
polyjdbc fix in seq generator

bartoszwalacik:
https://github.com/javers/javers/issues/475


bartoszwalacik:
waiting for central sync

Issue #455, opened 2016-11-11
MySQL error: Specified key was too long; max key length is 767 bytes


vrozkovec:
Hi, I'm initializing repository this way:
```
JaversSqlRepository sqlRepository = SqlRepositoryBuilder.sqlRepository()
	.withConnectionProvider(connectionProvider)
	.withDialect(DialectName.MYSQL)
	.build();
Javers javers = JaversBuilder.javers().registerJaversRepository(sqlRepository).build();
```
however I get an error in the application:

```
13:22:18,689  INFO SqlRepositoryBuilder:55 - starting up SQL repository module ...
13:22:18,719  INFO Java8AddOns:13 - loading Java8 add-ons ...
13:22:18,738  INFO ScannerModule:28 - using FIELD mappingStyle
13:22:18,753  INFO JaversSchemaManager:244 - creating javers table jv_commit_property ...
13:22:18,753  INFO SchemaManagerImpl:51 - creating entity with name jv_commit_property using ddl:
CREATE TABLE jv_commit_property (
property_name VARCHAR(200),
property_value VARCHAR(200),
commit_fk BIGINT,
CONSTRAINT jv_commit_property_pk PRIMARY KEY(commit_fk, property_name),
CONSTRAINT jv_commit_property_commit_fk FOREIGN KEY(commit_fk) REFERENCES jv_commit(commit_pk)
) ENGINE = InnoDB
13:22:18,755  WARN RequestCycleExtra:346 - ********************************
13:22:18,755  WARN RequestCycleExtra:347 - Handling the following exception
org.polyjdbc.core.exception.SchemaManagerException: [DDL_ERROR] Failed to run DDL:
CREATE TABLE jv_commit_property (
property_name VARCHAR(200),
property_value VARCHAR(200),
commit_fk BIGINT,
CONSTRAINT jv_commit_property_pk PRIMARY KEY(commit_fk, property_name),
CONSTRAINT jv_commit_property_commit_fk FOREIGN KEY(commit_fk) REFERENCES jv_commit(commit_pk)
) ENGINE = InnoDB
	at org.polyjdbc.core.schema.SchemaManagerImpl.ddl(SchemaManagerImpl.java:91)
	at org.polyjdbc.core.schema.SchemaManagerImpl.create(SchemaManagerImpl.java:52)
	at org.javers.repository.sql.schema.JaversSchemaManager.ensureTable(JaversSchemaManager.java:245)
[...]

PR #483, opened 2017-01-02
Fix for #455

raupachz:
This should fix the problem with the maximum key length in MySQL. Unfortunately I am not well versed with Gradle and fail to make this all work on my local machine.


bartoszwalacik:
this issue will be continued here https://github.com/javers/javers/pull/485/files
I've cleaned it up a bit

PR #495, opened 2017-01-26
Java8

bartoszwalacik:


LICENSE
```diff
@@ -175,9 +175,9 @@ Apache License
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
+   APPENDIX: How to test the Apache License to your work.
```
rafax: Looks like this got renamed accidentally?

bartoszwalacik: thanks, IDE has gone to far while refactoring

PR #545, opened 2017-06-04
Use CustomComparator for a object in list 492

pszymczyk:
It does not work wit Sets at the moment


javers-core/src/test/groovy/org/javers/core/CustomPropertyComparatorE2ETest.groovy
```diff
@@ -87,4 +88,26 @@ class CustomPropertyComparatorE2ETest extends Specification {
...
+        then:
+        javers.compare(left, right).changes.size() == 1
+
+        when:
+        javers = JaversBuilder.javers()
+                .registerValueChangeCustomComparator(new DummyCustomPropertyComparator(), String)
+                .build()
+
+        then:
+        javers.compare(left, right).changes.size() == 0
```
bartoszwalacik: So we only test that the new feature **doesn't** work?

javers-core/src/main/java/org/javers/core/JaversBuilder.java
```diff
@@ -477,6 +479,37 @@ public JaversBuilder withObjectAccessHook(ObjectAccessHook objectAccessHook) {
...
+     * Internally, given type is mapped as {@link CustomType}.
+     * <br/><br/>
+     *
+     * Custom types are serialized to JSON using Gson defaults.
+     *
+     * @param <T> custom type
+     * @param customType class literal to define a custom type
+     * @see CustomPropertyComparator
+     */
+    public <T> JaversBuilder registerValueChangeCustomComparator(CustomPropertyComparator<T, ValueChange> comparator, Class<T> customType){
```
bartoszwalacik: The new API is unclear for me. Why Custom Comparator for Collections of Values returns `ValueChange`?
When we compare Collections we produce  `ContainerChange` with list of entry changes (add, remove, etc).

Second, does it affect *normal* comparing of Values when they are simple properties and not elements of a Collection?

I think, that we should somehow extend the `CustomPropertyComparator` interface to support Collections,
and let user to register one CustomPropertyComparator per one Custom type

javers-core/src/main/java/org/javers/core/diff/custom/CustomValueComparator.java
```diff
@@ -0,0 +1,14 @@
...
+import org.javers.core.JaversBuilder;
+import org.javers.core.metamodel.type.ValueType;
+
+/**
+ * Can be used only for comparing {@link ValueType}.
+ *
+ * @see JaversBuilder#registerValue(Class, CustomValueComparator)
+ */
+@FunctionalInterface
+public interface CustomValueComparator<T> {
```
pszymczyk: very nice 👍

pszymczyk:
+1

Issue #548, opened 2017-06-19
PROPERTY_ACCESS_ERROR with Google AutoValue


jonfreedman:
Google have a nice library [AutoValue](https://github.com/google/auto/tree/master/value) to generate immutable object model classes.  To use this you create a public abstract class with abstract methods for each object property and at compile time an implementation is generated with package class access.

There are several issues when trying to use this library:

- The `Id` attribute is not picked up without switching to [MappingStyle.BEAN](http://javers.org/documentation/domain-configuration/#property-mapping-style)

- The `Id` attribute is not picked up on an abstract method

- Due to the way properties are looked up using reflection this results in `IllegalAccessException`s being throw from `org.javers.common.reflection.JaversGetter`.  It would be possible to avoid this & potentially other polymorphic issues by traversing up the class hierarchy until you find a public class on which to perform reflection.

I have created a failing test [here](https://github.com/jonfreedman/javers/blob/master/javers-core/src/test/java/org/javers/core/cases/AutoValueCase.java) which fails with the InvocationTargetException.  
If you move the `@Id` attribute from `Animal#getJaversId` to `Animal#getName` the tests fail with an `ENTITY_WITHOUT_ID` error.
If you remove `.withMappingStyle(MappingStyle.BEAN).registerEntity(Animal.class)` then the test will fail with id = 1.0 not Cat.

I would be happy to help fix this with some pointers.

jonfreedman:
I see that abstract methods are ignored in [JaversGetterFactory](https://github.com/javers/javers/blob/master/javers-core/src/main/java/org/javers/common/reflection/JaversGetterFactory.java#L45) - removing that requirement allows Animal#getName to be annotated with `@Id`.

bartoszwalacik:
Thanks for reporting. Well, we can do changes but we cant add explicit dependency on this google lib to javers-core. So is it possible to reword this issue in a more abstract way?

jonfreedman:
I added the dependency to the test-scope - I expect you could illustrate the issue with a non-abstract class that wasn't auto-generated.  I'm just attempting to dig into this further.

jonfreedman:
[...]

PR #549, opened 2017-06-19
Allow abstract methods to be used in JaversGetter

jonfreedman:
Regarding issue #548 these changes remove the restriction that methods cannot be abstract, this supports the situation where a concrete sub-class implements an abstract `@Id` annotated method, specifically for use with Google's AutoValue project.  In order to avoid breaking the existing TypeFactoryBeanIdTest JaversMember has a #looksLikeId property introduced which can be propagated from the sub-class to the super-class method to support the situation where a concrete sub-class implements a method adding the `@Id` annotation.


javers-core/src/test/groovy/org/javers/core/cases/AutoValueCase.groovy
```diff
@@ -0,0 +1,72 @@
...
+    diff.changes[0].left == 1
+    diff.changes[0].right == 2
+
+    first.author == 'Alice'
+    second.author == 'Bob'
+    second.changes.size() == 1
+    second.changes[0].propertyName == 'value'
+    second.changes[0].left == 1
+    second.changes[0].right == 2
+    second.snapshots[0].managedType.managedClass.looksLikeId.size() == looksLikeId
```
jonfreedman: I'm sure there is a better way to test this?

javers-core/src/test/groovy/org/javers/core/cases/AutoValueCase.groovy
```diff
@@ -0,0 +1,72 @@
...
+
+    where:
+    label << ['basic javers', 'javers with bean mapping', 'javers with bean mapping and entity registered']
+    javers << [
+            JaversBuilder.javers().build(),
+            JaversBuilder.javers().withMappingStyle(MappingStyle.BEAN).build(),
+            JaversBuilder.javers().withMappingStyle(MappingStyle.BEAN).registerEntity(AbstractEntity.class).build()
+    ]
+    looksLikeId << [0, 1, 1]
```
jonfreedman: The basic javers case can be removed, it's just here for illustration of why this whole change is necessary

jonfreedman:
Please let me know if you need anything else on this PR or if it can be included in 3.3.1

bartoszwalacik:
This code should be cleaned before merging, especially these multiply nested blocks in isOverriden(). Now i'm on devoxx, will back to you in a few days.

jonfreedman:
Sure, enjoy the conference, I'm developing against a local snapshot for the moment.

javers-core/src/main/java/org/javers/common/reflection/JaversGetterFactory.java
```diff
@@ -63,16 +69,35 @@ private static boolean isNotStatic(Method rawMethod) {
...
+
+    private static boolean isSubClass(final Method parent, final Method toCheck) {
+        return parent.getDeclaringClass().isAssignableFrom(toCheck.getDeclaringClass());
+    }
+
+    private static boolean sameMethodName(final Method parent, final Method toCheck) {
+        return parent.getName().equals(toCheck.getName());
+    }
+
+    private static boolean returnTypeCovariant(final Method parent, final Method toCheck) {
```
bartoszwalacik: +1 for taking care about return type covariance

PR #562, opened 2017-07-03
Correct data retrieval in CdoSnapshotMapper (TEXT columns)

grzegorz-chilkiewicz:
Data from TEXT/CLOB columns should be retrieved with getClob(), not with getString()


bartoszwalacik:
Why we need to change it? Can you write a test case which would show why the change is needed?

bartoszwalacik:
closing, because @grzegorz-chilkiewicz is not responding
You can come back with a failing test case which would show why the change is needed

PR #590, opened 2017-10-04
431 aspect around saveAndFlush

pszymczyk:


pszymczyk:
It needs some polishing.

Issue #605, opened 2017-11-21
Compare Lists Like Sets


drakin:
Add new compare Algorithm to be able compare lists like a sets (ignoring ordering).
For now, If we want to compare collection without ordering, we should to use sets.
In case, when we need to ignore ordering and we can not to change existing lists Objects to sets we have some troubles. 
We can not to change our model and use Sets.

bartoszwalacik:
👍

bartoszwalacik:
@drakin thanks for your contribution. Since merge, you are listed on https://javers.org/ as a JaVers contributor. Now we need to update the docs and the change will be ready to release.

bartoszwalacik:
I mean this section https://github.com/javers/javers.github.io/blob/master/documentation/diff-configuration/index.md

bartoszwalacik:
released in 3.7.0, see https://javers.org/documentation/diff-configuration/#list-algorithms

PR #606, opened 2017-11-21
#605 Add List compare Algorithm SET.

drakin:
Add new compare Algorithm to be able compare lists like a sets (ignoring ordering).
For now, If we want to compare collection without ordering, we should to use sets.
In case, when we need to ignore ordering and we can not to change existing lists Objects to sets we have some troubles. 
We can not to change our model and use Sets.

### DIFF
--- a/javers-core/src/test/groovy/org/javers/core/diff/appenders/SetListChangeAppenderTest.groovy
+++ b/javers-core/src/test/groovy/org/javers/core/diff/appenders/SetListChangeAppenderTest.groovy
@@ -1,96 +1,13 @@
 package org.javers.core.diff.appenders
-
-import org.javers.core.model.DummyUser
-import org.javers.core.model.SnapshotEntity
-import spock.lang.Unroll
-
-import java.time.LocalDate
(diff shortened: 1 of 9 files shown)


javers-core/src/test/groovy/org/javers/core/diff/appenders/SetListChangeAppenderTest.groovy
```diff
@@ -0,0 +1,96 @@
...
+
+import java.time.LocalDate
+
+import static org.javers.core.diff.appenders.ContainerChangeAssert.getAssertThat
+import static org.javers.core.model.DummyUser.dummyUser
+
+/**
+ * @author Sergey Kobyshev
+ */
+class SetListChangeAppenderTest extends AbstractDiffAppendersTest {
```
bartoszwalacik: it's a copy of SetChangeAppenderTest.
Both classes should share the same tests but different `setupSpec` (see BeanBasedScannerTest, FieldBasedScannerTest)

PR #615, opened 2017-12-03
Sketch for Whitelisting audited classes & properties 

bartoszwalacik:


bartoszwalacik:
see https://github.com/javers/javers/issues/616

bartoszwalacik:
thread moved here https://github.com/javers/javers/pull/628

PR #617, opened 2017-12-06
Mundodojava replace string id feature

bartoszwalacik:


mundodojava:
Thank you for opened a pull request :).

PR #621, opened 2017-12-18
Support for DB2 and DB2400

marvindaviddiaz:
https://github.com/polyjdbc/polyjdbc/pull/36

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
@@ -3,6 +3,7 @@ package org.javers.repository.sql.schema;
 import org.polyjdbc.core.dialect.Dialect;
 import org.polyjdbc.core.dialect.OracleDialect;
 import org.polyjdbc.core.dialect.MysqlDialect;
+import org.polyjdbc.core.schema.model.LongAttributeBuilder;
 import org.polyjdbc.core.schema.model.RelationBuilder;
 import org.polyjdbc.core.schema.model.Schema;
 import org.polyjdbc.core.util.StringUtils;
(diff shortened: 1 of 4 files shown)


bartoszwalacik:
why this is not connected with https://github.com/polyjdbc/polyjdbc/pull/36 ?

marvindaviddiaz:
Sorry I didn't understand your last comment. I think I should change this "polyjdbcVersion=0.7.1" but I'm not sure. Could you tell me what should I do? (I'm a beginner with gradle)

bartoszwalacik:
it's not about gradle, we can bump polyjdbcVersion in javers, but first, @adamdubiel need to release polyjdbc it to Central

bartoszwalacik:
OK, polyjdbc 0.7.2 is released, so now you can bump its version in javers and use the new dialects. I mean link javers dialect enum with polyjdbc dialect.

marvindaviddiaz:
Version was updated

javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
```diff
@@ -111,9 +112,9 @@ private Schema commitPropertiesTableSchema(Dialect dialect) {
         RelationBuilder relationBuilder = schema.addRelation(tableName.nameWithSchema());
         relationBuilder
             .primaryKey(tableName.localName() + "_pk").using(COMMIT_PROPERTY_COMMIT_FK, COMMIT_PROPERTY_NAME).and()
-            .withAttribute().string(COMMIT_PROPERTY_NAME).withMaxLength(MAX_INDEX_KEY_LEN_IN_MYSQL).and()
+            .withAttribute().string(COMMIT_PROPERTY_NAME).withMaxLength(MAX_INDEX_KEY_LEN_IN_MYSQL).notNull().and()
             .withAttribute().string(COMMIT_PROPERTY_VALUE).withMaxLength(600).and();
-        foreignKey(tableName, COMMIT_PROPERTY_COMMIT_FK, getCommitTableNameWithSchema(), COMMIT_PK, relationBuilder);
+        foreignKey(tableName, COMMIT_PROPERTY_COMMIT_FK, true, getCommitTableNameWithSchema(), COMMIT_PK, relationBuilder);
```
bartoszwalacik: why you want this additional not null constraint and why have you added this only for this fk?
SNAPSHOT_GLOBAL_ID_FK and SNAPSHOT_COMMIT_FK are alse not null

bartoszwalacik:
see my comment

javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
bartoszwalacik: I mean we have not null constraints mainly for primary keys so why to add not nulls for other columns?

marvindaviddiaz: I added that because COMMIT_PROPERTY_COMMIT_FK is part of a primary key. DB2 throws -542 error:

"-542   column-name CANNOT BE A COLUMN OF A HASH KEY, PRIMARY KEY, A UNIQUE CONSTRAINT, OR A PARENT KEY BECAUSE IT CAN CONTAIN NULL VALUES"

bartoszwalacik: ok, so `if Db2 ...` https://github.com/javers/javers/pull/624/commits/ce1f44a9e22b0aa04724b7a13e1ee7d38a5d5c99 ?

bartoszwalacik: fixed my typo, so final PR would be https://github.com/javers/javers/pull/624/files , right?

marvindaviddiaz: Yes, you're right

bartoszwalacik:
released in 3.7.8

Issue #519, opened 2017-03-13
Switch for enablig/disabling typeName (entity) index in MongoDB


croffler:
The mongoSchemaManager  drops the following index on startup:
dropIndexIfExists(snapshots, GLOBAL_ID_ENTITY);

To speed up some of my queries I created an index on globalId.entity.  Unfortunately every time I restart my app the index gets dropped.

Can this made configurable, to drop or not to drop ?

bartoszwalacik:
Are you sure that javers still uses this index?

bartoszwalacik:
I mean, this is only the name of a class, like `Person`. Index on this field wouldn't be *selective* in most cases and would be useful only when quering `byClass`. Do you have lot of different classes?

croffler:
Yes I have about 150 different entities and growing.

For example: 
List<CdoSnapshot> snapshots = getJavers().findSnapshots(QueryBuilder.byClass(klass).build()); 
 
JQL produces this:
{ $orderby: { commitMetadata.id: -1 }, $query: { $or: [ { globalId_key: /^rw\.gov\.ifmis\.payment\.domain\.PaymentOrder/.*/, globalId.entity: { $exists: true } } ] } }
You can now imagine that this is very slow with 20million snapshots(and growing) and 150 different classes. 

Here is my use case:
I want to find all snapshots  for a PaymentOrder with paymentOderNumber="xxx"

bartoszwalacik:
ok, so as a quick fix, lets remove this statement:

```
//schema migration script from JaVers 2.0 to 2.1
        dropIndexIfExists(snapshots, GLOBAL_ID_ENTITY);
```

 and you will be able to add index on your own

would you create a Pull Request?

croffler:
Hang on here ... there was a reason why thats implemented the way it is ...
Let me just create an index with a different name on that property in mongo .... and I should be ok.
In the meantime we have time to think about this a bit more .... perhaps this could be a configuration property that can be set with spring property injection ....(dropIndexes="${javers.dropIndexes}")

bartoszwalacik:
ok, switch would be better. I would call it `useTypeNameIndex`, with default value false

croffler:
sounds good

bartoszwalacik:
released in 3.7.7

PR #622, opened 2017-12-19
#519 implemented typeName index tumbler

tau3:
it's ready; one thing that bothers me is MongoRepository class - it has like 5 constructors.
probably it should have one constructor and a builder (which can be a subject of another issue)


bartoszwalacik:
I'm confused, why so many changes here?
The issue you have linked is about Switch for enablig/disabling typeName.

tau3:
had to refactor MongoSchemaManager to make it testable.
otherwise the test would be too complicated; a complicated test usualy means that tested code is complicated too)

bartoszwalacik:
@tau3 leave it, I have simply added this index https://github.com/javers/javers/commit/e5fc200a01828f63f3270efdb7bd2f23dcd55039

PR #628, opened 2018-01-14
Properties whitelisting

bartoszwalacik:


stefanicai:
@bartoszwalacik  Looks good to me. Maybe we can add in the javadocs of ClientsClassDefinitionBuilder that only one (whitelisted or ignored) can be specified. They would get an exception if they specify both, but maybe it's nicer if they can read it  before?

stefanicai:
Is there a reason why you named that property whitelisted instead of included? That seems less intuitive to me and a few friends I ask about. Ideally it'd have been included/excluded.

stefanicai:
There's another requirement there, the whitelisted props should be applied to subclasses too (same as ignored props). So a subclass should only compare/log those specific props unless we explicitly provide an EntityDefinition for that subclass.

bartoszwalacik:
@stefanicai 
- yes we need to deal with inheritance, please commit this test, then we will implement ignoring all props in subclases
- you are right about javadoc, please add id
- I wouldn'd argue about names, if you guys are native speakers, feel free to change back to Included, but Ignored have to stay, we don't want to change existing API

stefanicai:
@bartoszwalacik please have a look at the updated version. I've added the failing test here, will add the annotation shortly.

bartoszwalacik:
Ok, I will look at this tomorrow morning

PR #631, opened 2018-01-23
DiffInclude annotation

stefanicai:
@bartoszwalacik here's a draft, to get an idea where I'm going with this. I can finish it once you answer the question on Gitter.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/annotation/DiffInclude.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/annotation/DiffInclude.java
@@ -7,12 +7,16 @@ import static java.lang.annotation.ElementType.*;
 import static java.lang.annotation.RetentionPolicy.RUNTIME;
 
 /**
- * Use DiffInclude annotation to tell JaVers what properties to include in the diff results for this type.
- * All other properties in the current type and it's subclasses will be ignored by JaVers.
+ * Use {@code DiffInclude} annotation to tell JaVers which properties to include in diff/commit
+ * operations for a given class.
+ * All other properties in this class and all properties in its subclasses will be ignored by JaVers.
+ * <br/>
  *
- * If more or different properties should be included in a subclass, apply the annotation to the subclass with the different list of properties.
+ * If some properties in a subclass should be included, apply the {@code DiffInclude} annotation on them.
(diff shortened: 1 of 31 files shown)


bartoszwalacik:
ok, I will take a look

javers-core/src/main/java/org/javers/core/metamodel/annotation/DiffInclude.java
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * class B extends A {
+ *     &#64;DiffIgnore
+ *     private String qux;
+ * }
+ *
+ * @see DiffIgnore
+ * @author Iulian Stefanica
+ */
+@Target({ TYPE})
```
bartoszwalacik: `@Target({METHOD, FIELD})` ?

javers-core/src/main/java/org/javers/core/metamodel/type/TypeFactory.java
```diff
@@ -42,23 +43,19 @@ JaversType create(ClientsClassDefinition def) {
     JaversType create(ClientsClassDefinition def, ClassScan scan) {
         if (def instanceof CustomDefinition) {
             return new CustomType(def.getBaseJavaClass());
-        } else
-        if (def instanceof EntityDefinition) {
+        } else if (def instanceof EntityDefinition) {
             return entityTypeFactory.createEntity((EntityDefinition) def, scan);
-        } else
```
bartoszwalacik: could you please redo the reformatting? it makes code review harder

bartoszwalacik:
I think we have it, what do you think, is it ok?

stefanicai:
Think so, just wanted to add another test, but I'm travelling tomorrow and Tuesday.

javers-core/src/main/java/org/javers/core/metamodel/type/TypeFactory.java
stefanicai: Do you have a formatting file that I can import in IntelliJ so I format code the same as you do?

bartoszwalacik:
we need one more test for this case:

If some properties in a subclass should be included, apply the {@code DiffInclude} annotation on them.

Issue #542, opened 2017-05-18
Possibility to manage Javers' SQL schema creation


bartoszwalacik:
- flag to disable automatic schema creation
- method to call schema creation programatically

antoniocasino:
Hi , I wrote a code that could manage schema creation, I wrote a test and connected it with h2. I had some issue with dependencies ( it's the first time I use gradle ) but it seems to work correctly.
I cloned the repo. What is the best way to share it?

bartoszwalacik:
hi, just fork javers repo and then create the Pull Request

bartoszwalacik:
see https://github.com/antoniocasino/javers/blob/master/CONTRIBUTING.md

bartoszwalacik:
@antoniocasino, just released in 3.8.1

PR #637, opened 2018-02-13
 Possibility to manage Javers SQL schema creation

antoniocasino:
I implemented it with a custom schema manager that can be injected in JaversSqlRepository


javers-persistence-sql/src/test/java/org/javers/repository/sql/CustomSchemaManager.java
```diff
@@ -0,0 +1,20 @@
...
+import org.javers.repository.sql.schema.SchemaNameAware;
+import org.javers.repository.sql.schema.TableNameProvider;
+
+import javax.swing.text.html.Option;
+import java.util.Optional;
+
+/**
+ * Created by antonio on 12/02/2018.
+ */
+public class CustomSchemaManager extends SchemaNameAware {
```
bartoszwalacik: This class does nothing, I don't get it

javers-persistence-sql/src/test/java/org/javers/repository/sql/cases/Case542CustomSchemaMaager.java
```diff
@@ -0,0 +1,60 @@
...
+    @Before
+    public void shouldCreateSchema(){
+        JaversBuilder.javers().registerJaversRepository(H2RepositoryFactory.create()).build();
+    }
+
+    @Test
+    public void shouldCommitEntityWithSerializableId() {
+        //given
+        Javers javers = JaversBuilder.javers()
+                .registerJaversRepository(H2RepositoryFactory.createWithCustomSchemaManager()).build();
```
bartoszwalacik: Since CustomSchemaManager does nothing, what is the purpose of this test? What feature are you testing here?

javers-persistence-sql/src/main/java/org/javers/repository/sql/JaversSqlRepository.java
```diff
@@ -46,6 +47,11 @@ public JaversSqlRepository(PolyJDBC polyJDBC, CommitMetadataRepository commitRep
         this.sqlRepositoryConfiguration = sqlRepositoryConfiguration;
     }
 
+    public JaversSqlRepository withSchemaManager(SchemaNameAware schemaManager){
```
bartoszwalacik: the issue https://github.com/javers/javers/issues/542 is about two things:

- flag to disable automatic schema creation
- method to call schema creation programatically

So please implement:
- a flag method - `SqlRepositoryBuilder.withSchemaManagementEnabled(boolean)`

`JaversSqlRepository.ensureSchema()` is already here, so the second point is already done

bartoszwalacik: I mean, we don't want CustomSchemaManager

javers-persistence-sql/src/test/java/org/javers/repository/sql/cases/Case542CustomSchemaMaager.java
antoniocasino: even if custom schema manager does not create tables it still write on them, indeed data is written via javers commit. CustomSchemaManager use another change I introduced that is the ability to instance TableNameProvider with an optional instead of using SqlRepositoryConfiguration that has a package visibility

javers-persistence-sql/src/main/java/org/javers/repository/sql/JaversSqlRepository.java
antoniocasino: Ok, 
so do you suggest to check if we want schema management with a flag and if it is true then invoke ensureSchema()?

bartoszwalacik: yes, and the flag should be set to true by default. Also, this flag should be available in Spring boot starters. see https://javers.org/documentation/spring-boot-integration/#javers-configuration-properties

antoniocasino: I implemented the feature but only tested manually ( changing default value of JaversProperties to false and see the exception). Could you give me a clue on how to test it with spring boot?

javers-persistence-sql/src/main/java/org/javers/repository/sql/pico/SchemaManagementDisabledModule.java
```diff
@@ -0,0 +1,40 @@
...
+import org.javers.repository.sql.schema.TableNameProvider;
+import org.polyjdbc.core.query.QueryRunnerFactory;
+
+import java.util.Arrays;
+import java.util.Collection;
+
+/**
+ * Created by antonio on 13/02/2018.
+ */
+public class SchemaManagementDisabledModule implements JaversModule {
```
bartoszwalacik: you don't need to copy-paste the whole class, one `if` in `JaversSqlRepository` would do the job:


```
    @Override
    public void ensureSchema() {
        if (... check the new flag) {
            schemaManager.ensureSchema();
        }
    }
```

bartoszwalacik: At the end, please squash your commits

antoniocasino: you told me to implement
 - SqlRepositoryBuilder.withSchemaManagementEnabled(boolean)
Do you mean that I have to pass this data to JaversSqlRepository?

bartoszwalacik:
ok, last thing, could you please squash your commits? (there are 9 of them)

antoniocasino:
squash done

PR #646, opened 2018-03-01
Ismaelgomescosta ismaelgomescosta entity as

bartoszwalacik:


javers-core/src/test/groovy/org/javers/core/JaversDiffE2ETest.groovy
```diff
@@ -114,6 +114,35 @@ class JaversDiffE2ETest extends AbstractDiffTest {
...
+
+    class EntityAsId {
+        @Id
+        int id
+        int value
+    }
+
+    def "should use nested Entity Id value as Id of parent Entity"(){
+        given:
+        def javers = javers().build()
```
bartoszwalacik: When you don't register EntityAsId as Entity, this test fails, because
EntityAsId is interred as Value. and when you don't touch DummyWithEntityId, EntityAsId would be inferred as Entity. This is not ok.

see how TypeFactory.inferIdPropertyTypeAsValue() works

bartoszwalacik: okay, I've fixed this issue in ` TypeFactory.inferIdPropertyTypeAsValue()`, also I've simplified the change in `localIdAsString()`

ismaelgomescosta: I liked your changes. I didn't know about the core of Javers. I will study more about this. Thank you.

Issue #677, opened 2018-05-10
Support current versions of Spring Data (2.x)


shollander:
Current versions of Spring Data use `saveAll(Iterable)` instead of `save(Iterable)`. When used with `JaversSpringDataAuditableRepositoryAspect`, calls to `saveAll()` are ignored. I believe this is a simple fix that just requires the addition of another pointcut definition in `JaversSpringDataAuditableRepositoryAspect`. `delete(Iterable)` has also been renamed to `deleteAll(Iterable)`.

See [DATACMNS-944](https://jira.spring.io/browse/DATACMNS-944) and [727ab8384cac21e574526f325e19d6af92b8c8df](https://github.com/spring-projects/spring-data-commons/commit/727ab8384cac21e574526f325e19d6af92b8c8df).

bartoszwalacik:
released in 3.9.7

PR #679, opened 2018-05-14
#677 Support Spring Data 2.x

shollander:


bartoszwalacik:
nice, but maybe some tests? 
see https://github.com/javers/javers/blob/master/javers-spring/src/test/groovy/org/javers/spring/auditable/integration/JaversSpringDataAspectIntegrationTest.groovy

shollander:
@bartoszwalacik I was looking for an existing unit test to update but I missed that one since it does not belong to the same package as the aspect class.
However I believe updating the unit test would require a compile-time dependency on the newer version of spring data commons, which would mean updating the root `gradle.properties` file. Is that what you want? That would mean, among other things, that the existing unit tests for `save(Iterable)` would become invalid.

bartoszwalacik:
Could be a test only dependency, but if save(iterable) is removed it indeed could break existing tests. So looks like we have to release your change without tests

Issue #682, opened 2018-06-21
Potential bug due to the non-determinicity of Class#getDeclaredConstructors


Oja95:
During initializing the modules, reflection is used to instantiate various classes needed for the modules.
However, the utility responsible for instantiating these classes takes the first public constructor from the array returned by `Class#getDeclaredConstructors` method and uses it to instantiate the class. See `org.javers.common.reflection.ReflectionUtil#newInstance`.

However, the Javadoc from `Class#getDeclaredConstructors` states the following:
> The elements in the array returned are not sorted and are not in any particular order.

Thus, relying on the order of the constructors obtained might result in unexpected behaviour.

A real erroneous scenario occurred during initializing the `InMemoryRepositoryModule`, more specifically, during instantiating the `InMemoryRepository` class. `InMemoryRepository` has two public constructors:

`public InMemoryRepository()`
`public InMemoryRepository(CommitIdGenerator commitIdGenerator)`

Should the `Class#getDeclaredConstructors` return an array in which the default public constructor is the first element, everything works well. However, should the second constructor be the first element, then it fails to resolve the `CommitIdGenerator` object reference from the Pico container. The following is then logged:

```
ERROR org.javers.core.pico.ContainerArgumentResolver - failed to get component of type org.javers.core.CommitIdGenerator from Pico container
ERROR org.javers.common.reflection.ReflectionUtil - failed to create new instance of org.javers.repository.inmemory.InMemoryRepository, argument resolver for arg[0] org.javers.core.CommitIdGenerator thrown exception: COMPONENT_NOT_FOUND: JaVers bootstrap error - component of type 'org.javers.core.CommitIdGenerator' not found in container
```

#### Steps to reproduce
There is no good way to reliably reproduce this by some simple application. To reproduce the issue, create a simple `main` method that builds a default Javers instance using the following:

```
JaversBuilder.javers().build();
```

Then add a conditional breakpoint to line 70 in `org.javers.common.reflection.ReflectionUtil` to only break when `clazz == InMemoryRepository.class` holds. Run the application, and when the application stops on the breakpoint, check the `constructor` variable to see if it references the default constructor. If so, evaluate an expression to change the value of the constructor variable to become the other constructor obtained by `Class#getDeclaredConstructors` using the following expression:

```
constructor = clazz.getDeclaredConstructors()[1];
```

[...]

PR #683, opened 2018-06-21
#682 Fix potential bug due to Class#getDeclaredConstructors non-determinicity

Oja95:
A potential solution for the #682 issue.
It tries all the public constructors before failing.

### DIFF
--- a/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
+++ b/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
@@ -5,9 +5,7 @@ import org.javers.common.collections.Lists;
 
 import java.util.*;
 
-import org.javers.common.collections.Primitives;
 import org.javers.common.collections.Sets;
-import org.javers.common.collections.WellKnownValueTypes;
 import org.javers.common.exception.JaversException;
 import org.javers.common.exception.JaversExceptionCode;
 import org.javers.common.validation.Validate;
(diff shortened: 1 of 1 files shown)


javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
```diff
@@ -62,25 +60,22 @@ public static Object invokeGetter(Object target, String getterName) {
...
             Object[] params = new Object[types.length];
-            for (int i=0; i<types.length; i++){
-                try {
-                    params[i] = resolver.resolve(types[i]);
-                } catch (JaversException e){
-                    logger.error("failed to create new instance of "+clazz.getName()+", argument resolver for arg["+i+"] " +
-                                 types[i].getName() + " thrown exception: "+e.getMessage());
-                    throw e;
-                }
+            try {
```
bartoszwalacik: to be honest, i dont know what did change in logic of this method (besides refactoring),
could you please explain the rationale behind this change?

Oja95: I moved the loop that resolved the constructor's parameters into a separate method. The main idea was that, if the constructor parameters resolving fails, the exception is stored in the `ex` variable (https://github.com/javers/javers/pull/683/commits/0f43b6cd37f170f8727a9f2ddc69b0773ae32950#diff-1e490987d4273f55c02bf8d2abf7a6adR66) and next public constructor is then tried. Should other constructors also fail, the exception is thrown at the end. If any of the other constructors succeed in instantiating the class object then the exception won't be thrown.

Oja95:
Would like to add that this might not be the best solution, it'll work around the described issue (#682) but also introduces another potentially problematic scenario. Since it iteratively looks through the array provided by `Class#getDeclaredConstructors`, it is not deterministic which constructor eventually will be used if multiple constructor candidates could instantiate the class.

There are some other approaches to tackle the problem:
1. Explicitly specify which constructor to use and obtain the concrete constructor reference via  `Class#getDeclaredConstructor`.
2. Always try default constructor first, then others (not sure if semantically correct in the framework's context)

bartoszwalacik:
fixed in https://github.com/javers/javers/pull/686/files

PR #700, opened 2018-08-15
javersexception-property-not-found

bartoszwalacik:
see https://stackoverflow.com/questions/51634751/javersexception-property-not-found-property-in-derived-class-not-found-in-abstr


bartoszwalacik:
released in 3.11.3

PR #713, opened 2018-09-16
Issue-712

vtiwari2134:
Added test case to reproduce issue while saving entity with javers auditing enabled


bartoszwalacik:
moved here https://github.com/javers/javers/pull/716

Issue #714, opened 2018-09-17
Cannot register custom comparator for List


kucharzyk:
I want to override javers collection comparator.

I've tried to register custom comparator for List.class like this:

```
TransactionalJaversBuilder
            .javers()
            .withTxManager(transactionManager)
            .registerJaversRepository(sqlRepository)
            .withObjectAccessHook(new HibernateUnproxyObjectAccessHook())
            .withProperties(javersSqlProperties)
            .withListCompareAlgorithm(ListCompareAlgorithm.LEVENSHTEIN_DISTANCE)
            .registerCustomComparator((left, right, affectedId, property) -> {
                System.out.println("WHATEVER");
                return null;
            }, List.class)
            .build();
```

Unfortunatelly I've got the following exception (my comparator is never invoked - exception is thrown before)

`
java.lang.ArrayIndexOutOfBoundsException: 0

	at com.google.gson.internal.$Gson$Types.resolveTypeVariable($Gson$Types.java:430)
	at com.google.gson.internal.$Gson$Types.resolve($Gson$Types.java:352)
	at com.google.gson.internal.$Gson$Types.resolve($Gson$Types.java:381)
	at com.google.gson.internal.$Gson$Types.resolve($Gson$Types.java:337)
	at com.google.gson.internal.$Gson$Types.getSupertype($Gson$Types.java:283)
	at com.google.gson.internal.$Gson$Types.getCollectionElementType($Gson$Types.java:302)
	at com.google.gson.internal.bind.CollectionTypeAdapterFactory.create(CollectionTypeAdapterFactory.java:52)
	at com.google.gson.Gson.getAdapter(Gson.java:458)
	at com.google.gson.Gson.fromJson(Gson.java:926)
	at com.google.gson.Gson.fromJson(Gson.java:994)
	at com.google.gson.internal.bind.TreeTypeAdapter$GsonContextImpl.deserialize(TreeTypeAdapter.java:162)
	at org.javers.core.json.typeadapter.commit.CdoSnapshotStateDeserializer.decodePropertyValue(CdoSnapshotStateDeserializer.java:66)
	at org.javers.core.json.typeadapter.commit.CdoSnapshotStateDeserializer.lambda$deserialize$0(CdoSnapshotStateDeserializer.java:38)
	at java.base/java.util.Iterator.forEachRemaining(Iterator.java:133)
	at java.base/java.util.Spliterators$IteratorSpliterator.forEachRemaining(Spliterators.java:1801)
[...]

PR #715, opened 2018-09-17
test case for issue #714

kucharzyk:
test case for issue #714


bartoszwalacik:
see comments below https://github.com/javers/javers/issues/714

Issue #712, opened 2018-09-15
Error while saving composite entity having another entity id as null


vtiwari2134:
Following exception occurs while saving entity with javers auditing enabled in spring - 
JaversException ENTITY_INSTANCE_WITH_NULL_ID: Found Entity instance 'com.test.entity.DepartmentEntity' with null idProperty 'id'

I think the issue is with <code>JaversSpringDataAuditableRepositoryAspect.class</code> having aspect on save (<code>@AfterReturning("execution(public * save(..)) && this(org.springframework.data.repository.CrudRepository)")</code>) method. In this the entity is audited on the argument passed to save method but auditing should be performed on return entity that is saved by hibernate.

Issue arises because whenever detached entity is saved to persist hibernate tries to merge the entity creating new entity in hibernate session. Hence the return entity is different from the entity passed to save.
That why the Aspect <code>public void onSaveExecuted(JoinPoint pjp) {
    this.onVersionEvent(pjp, this.saveHandler);
  }</code> should evaluate the return entity from <code>save</code> method.

**Sample spring project to replicate bug -**
[test.zip](https://github.com/javers/javers/files/2385977/test.zip)

To replicate issue run JaversExceptionTest  test case. This test class contains 2 test one saves entity without having auditing enabled and one with javers auditing enabled.
In first case EmployeeEntity is saved successfully in db while second case throws error : -

<code>JaversException ENTITY_INSTANCE_WITH_NULL_ID: Found Entity instance 'com.test.entity.DepartmentEntity' with null idProperty 'id'

	at org.javers.core.metamodel.type.EntityType.getIdOf(EntityType.java:88)
	at org.javers.core.metamodel.object.InstanceId.createFromInstance(InstanceId.java:28)
	at org.javers.core.metamodel.object.GlobalIdFactory.createId(GlobalIdFactory.java:49)
	at org.javers.core.graph.LiveCdoFactory.create(LiveCdoFactory.java:25)
	at org.javers.core.graph.LiveCdoFactory.create(LiveCdoFactory.java:10)
	at org.javers.core.graph.EdgeBuilder.buildSingleEdge(EdgeBuilder.java:30)</code>

bartoszwalacik:
@vtiwari2134 could you please push this test case to github? see https://github.com/javers/javers/tree/master/javers-spring-boot-starter-sql/src/test/groovy/org/javers/spring/sql

vtiwari2134:
[...]

PR #718, opened 2018-09-19
Fix : Checking version on saved entity rather than on the passed parameter

vtiwari2134:
As per comment : https://github.com/javers/javers/issues/712#issuecomment-422201793


vtiwari2134:
Creating new PR with all test cases covered.

PR #719, opened 2018-09-23
Fix for Issue 712

vtiwari2134:
1.) Enable auditing on persisted object rather than the entity passed as parameter
2.) Resolved one more issue with lazy hibernate proxy object (Added inline description with tag Comment-101 in code)

Test class - ObjectAccessHookFieldSpec.groovy

Screenshot for hibernate proxy object test -
persontable
jv_snapshot


bartoszwalacik:
When I run this branch on local, some tests fail. I have pushed this branch to javers repo and the build also fails. I need to figure out what's going on because your changes seems right.

Issue #723, opened 2018-10-02
How to handle removed entity field?


chengchen:
Hello,

I am sorry if this question is already asked, and I did search before asking, the most relevant one I found is: https://github.com/javers/javers/issues/511

We are using javers to do audits in our company, and we are satisfied with it's simplicity. However, we had an issue recently regarding a simple removal of a field in an Hibernate entity. And this field itself is an entity as well. So to simplify a bit: Product entity had a field called Category, but Category has been removed completely from db as well as from the code. If I don't do any change in the javers tables, it will break simply with the exception:

```
org.javers.common.exception.JaversException: TYPE_NAME_NOT_FOUND: type name 'com.edgelab.marketdata.domain.Issuer' not found. If you are using @TypeName annotation, remember to register this class using JaversBuilder.withPackagesToScan(String) or JaversBuilder.scanTypeName(Class)
	at org.javers.core.metamodel.type.TypeMapperState.getClassByDuckType(TypeMapperState.java:70) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.metamodel.type.TypeMapperState.getClassByTypeName(TypeMapperState.java:42) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.metamodel.type.TypeMapper.getJaversManagedType(TypeMapper.java:168) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.json.typeadapter.commit.GlobalIdTypeAdapter.parseEntity(GlobalIdTypeAdapter.java:117) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.json.typeadapter.commit.GlobalIdTypeAdapter.parseInstanceId(GlobalIdTypeAdapter.java:59) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.json.typeadapter.commit.GlobalIdTypeAdapter.fromJson(GlobalIdTypeAdapter.java:36) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.json.typeadapter.commit.GlobalIdTypeAdapter.fromJson(GlobalIdTypeAdapter.java:15) ~[javers-core-3.11.5.jar:na]
	at org.javers.core.json.JsonConverterBuilder.lambda$registerJsonTypeAdapterForType$4(JsonConverterBuilder.java:168) ~[javers-core-3.11.5.jar:na]
....
```

[...]

PR #726, opened 2018-10-08
#723 handle removed entity by registering a dummy entity

chengchen:
Hey @bartoszwalacik, I tried to make the tests pass but I am not sure about the implementation if it fits your standard, since I am not fully aware of the whole picture of javers.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapper.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapper.java
@@ -172,6 +172,33 @@ public class TypeMapper {
 
     /**
      * @throws JaversException TYPE_NAME_NOT_FOUND if given typeName is not registered
+     * @since 3.11.7
+     */
+    public <T extends ManagedType> T getJaversManagedTypeIncludingRemoved(String typeName, Class<T> expectedType) {
+        try {
+            return getJaversManagedType(typeName, expectedType);
+        } catch (JaversException e) {
+            if (e.getCode() == JaversExceptionCode.TYPE_NAME_NOT_FOUND) {
+                // the entity type might have been removed, try to register a RemovedEntity for it
+                registerClientsRemovedType(typeName);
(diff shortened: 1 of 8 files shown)


bartoszwalacik:
i will take a look, but looks like this issue is more complicated than I thought

chengchen:
@bartoszwalacik After testing the integrations with our applications, it seems that it's not enough the spec... To retrieve the owner snapshots, it's still not working (but the changes should be similar). I am rejecting the PR.

javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapper.java
```diff
@@ -168,6 +170,33 @@ public ManagedType getJaversManagedType(GlobalId globalId){
         return getJaversManagedType(state.getClassByTypeName(typeName), expectedType);
     }
 
+    /**
+     * @throws JaversException TYPE_NAME_NOT_FOUND if given typeName is not registered
```
chengchen: TYPE_NAME_NOT_FOUND is the only exception here? If yes, then the comment needs to be refined.

bartoszwalacik:
@chengchen finally, I have implemenented it myself, see https://github.com/javers/javers/pull/725
and I'd like to merge my version.
Your version works but `RemovedEntity.java` would be problematic. Your work was very valuable, it helped me to solve this problem.

You are encouraged to contribute more to javers.

chengchen:
@bartoszwalacik no worries. As long as I helped :D
Care to tell why RemovedEntity is problematic here? I would like to understand as well.

bartoszwalacik:
Javers uses information about property name and type to deserialize Snapshots. 
For example, if your class has Id field of type `int`, value of this field in deserialized Snapshot would be also `int` type. If this class is not found, JSON basic types are used, so the type of this value would be `double`. In this context, it's better to have no type info than wrong type info (like in fake RemovedEntity class).

Issue #690, opened 2018-07-16
Missing property in *.yml to set SQL database schema name in javers-spring-boot-starter-sql


bartoszwalacik:
props should be added here

https://github.com/javers/javers/blob/master/javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSqlAutoConfiguration.java#L67

bartoszwalacik:
released in 3.12.4

PR #734, opened 2018-10-26
#690 Missing property in *.yml to set SQL database schema name in jav…

MrMangoNerd:


javers-spring-boot-starter-sql/src/test/resources/application-test.yml
```diff
@@ -5,6 +5,7 @@ javers:
   prettyPrint: false
   typeSafeValues: true
   dialect: H2
+  schema: 'test'
```
bartoszwalacik: nice PR
```suggestion
  schema: 'test' please change this prop name to sqlSchema or sqlSchemaName
```

PR #754, opened 2018-11-27
Add LGTM code quality badges

xcorail:
Hi there!

I thought you might be interested in adding these LGTM [code quality badges](https://lgtm.com/projects/g/javers/javers/ci/#badges) to your project. They indicate how you care about code quality and encourage your future contributors to do the same.
To get an idea of the analyses reflected by these grades, check the [alerts discovered by LGTM](https://lgtm.com/projects/g/javers/javers).

N.B.: I am on the team behind LGTM.com, I'd appreciate your feedback on this initiative, whether you're interested or not, if you find time to drop me a line. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,8 @@
 
 [![Maven Central](https://maven-badges.herokuapp.com/maven-central/org.javers/javers-core/badge.svg)](https://maven-badges.herokuapp.com/maven-central/org.javers/javers-core)<br/>
 [![Build Status](https://travis-ci.org/javers/javers.svg?branch=master)](https://travis-ci.org/javers/javers/)
+[![Code Quality: Java](https://img.shields.io/lgtm/grade/java/g/javers/javers.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/javers/javers/context:java)
+[![Total Alerts](https://img.shields.io/lgtm/alerts/g/javers/javers.svg?logo=lgtm&logoWidth=18)](https://lgtm.com/projects/g/javers/javers/alerts)
 
 ## What is JaVers
 


bartoszwalacik:
Hi, nice idea, but looks like your framework fires some trivial false positives, for example.

> This ResultSet is not always closed on method exit.

as javadoc ResultSet.close() says:

Note: A ResultSet object is automatically closed by the Statement object that generated it when that Statement object is closed, re-executed, or is used to retrieve the next result from a sequence of multiple results.

xcorail:
Hello @bartoszwalacik 

Thanks for the feedback. 
I will have a closer look at the FP you reported.

Can you please point me to the other trivial FPs you noticed, if you have time?

Cheers

bartoszwalacik:
for example this one
https://lgtm.com/projects/g/javers/javers/snapshot/156d22c4c1d22229235ca31e0dd278b9f19b71c6/files/javers-core/src/main/java/org/javers/core/JaversBuilder.java?sort=name&dir=ASC&mode=heatmap#x9705087b258a59f:1

It's the first 'error' on your list, but it's just an unused variable, the minor issue, definitely not an error.

bartoszwalacik:
this one has value for me
https://lgtm.com/projects/g/javers/javers/snapshot/156d22c4c1d22229235ca31e0dd278b9f19b71c6/files/javers-core/src/main/java/org/javers/core/metamodel/type/TypeMapperState.java?sort=name&dir=ASC&mode=heatmap#x9d0aa98906cd2819:1

bartoszwalacik:
ok then, please let me know when you fix this ResultSet FP, then I reconsider merge,
bests

xcorail:
Hi @bartoszwalacik 

After a debate within the team, they prefer to keep the query as it is. They agree that in your particular case it's a FP, but in other cases, if the Statement object generating the ResultSet is not closed it would be a TP. 
They also think that in general it's a good practice to use TryWithResources, so they don't want to make an exception to this rule for ResultSet. 

Cheers
Xavier

bartoszwalacik:
I mean, when you check if statement are closed, checking if result sets are closed is pointless, since a statement closes its child result sets.

PR #772, opened 2019-01-16
Instant serialization fix

mwesolowski:
In JDK 11 java.time.Instant has microseconds precision.


bartoszwalacik:
Seems legit but we need to extend the COMMIT_COMMIT_DATE_INSTANT col in SQL repo.
Now it has size 26, we need more to store microseconds.
let's continue work here https://github.com/javers/javers/pull/776

PR #776, opened 2019-01-20
Mwesolowski instant serialization

bartoszwalacik:


bartoszwalacik:
My intentions is :
1. do not change the precision of `commitDate`, for the following reasons:
- troublesome data and SQL schema migration
- we are using native SQL type (timestamp), some databases have timestamp with micro precision and some doesn't have
- leave it as Local timestamp (backward compatibility) 

2. Provide the new, reliable (time zone aware) and precise timestamp as `Instant commitDateInstant` mapped to `jv_snapshot.commit_date_instant`, with precision up to micros (depending on the jdk)

So I think, in this PR we shoudn't touch `commitDate` and focus on making `commitDateInstant` work on 9 and jdk with `Instant`  with micro precision

javers-core/src/test/groovy/org/javers/core/JaversRepositoryE2ETest.groovy
```diff
@@ -85,17 +85,14 @@ class JaversRepositoryE2ETest extends Specification {
...
 
         when:
         javers.commit('author', new SnapshotEntity(id: 1))
         def snapshot = javers.getLatestSnapshot(1, SnapshotEntity).get()
 
         then:
-        println 'commitDate: ' + snapshot.commitMetadata.commitDate
-        println 'commitDateInstant: ' + snapshot.commitMetadata.commitDateInstant
-        snapshot.commitMetadata.commitDate == now.toLocalDateTime()
+        ChronoUnit.MILLIS.between(snapshot.commitMetadata.commitDate, now.toLocalDateTime()) == 0
```
mwesolowski: It turned out that Spock test cannot be easily overridden in a subclass, therefore I have to introduce this change here and not in MySqlIntegrationTest.

mwesolowski:
Current version of this PR is here: https://github.com/javers/javers/pull/777

PR #779, opened 2019-02-01
Sorted set bug test case

ammachado:
Client domain object snapshots are unable to deserialize, due the fact that GlobalId instances are not `Comparable`.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/metamodel/object/GlobalId.java
+++ b/javers-core/src/main/java/org/javers/core/metamodel/object/GlobalId.java
@@ -1,5 +1,6 @@
 package org.javers.core.metamodel.object;
 
+import java.util.Comparator;
 import org.javers.common.validation.Validate;
 import org.javers.core.metamodel.type.ManagedType;
 import org.javers.repository.jql.GlobalIdDTO;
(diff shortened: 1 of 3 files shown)


ammachado:
It seems that the build is breaking because of Java 11, not because the patch itself.

bartoszwalacik:
@ammachado could you please explain your intentions? Why `GlobalId` should be `Comparable` and how it's linked with `SortedSet`?

ammachado:
@bartoszwalacik, it is because JaVers (or Gson, which is responsible for snapshot serialisation/deserialisation in this case) seems to reuse the collection type of the entity being audited (please, see attached test case on this PR).

When using a normal `Set`, JaVers can find and deserialise the snapshots correctly. When using a `SortedSet`, it can't, because the `SortedSet`'s object must be comparable, given that no `Comparator<E>` instance can be supplied during the process.

Implementing `Comparable` on `GlobalId` was the naivest solution.  Perhaps you can help finding a better solution.

bartoszwalacik:
thanks, I will merge

bartoszwalacik:
that's crazy, after this change, some tests fails on openjdk 11.0.2 but on 11.0.1 all are green

bartoszwalacik:
Tests are failing due to Groovy assertion issue, I will fix it.

Issue #788, opened 2019-02-18
MongoDB schema creation fails in AWS DocumentDB due to overly-long index name


jpars:
Hello,

When using `MongoRepository` with AWS DocumentDB, `MongoSchemaManager.ensureSchema()` fails to create the compound index of `commitMetadata.properties.key` and `commitMetadata.properties.value` in jv_snapshots, because the autogenerated index name is too long. Both MongoDB and DocumentDB limit the fully-qualified index name (`<db>.<col>.$<index>`) to under 128 characters, but DocumentDB places the additional restriction that the unqualified index name (`<col>.$<index>`) must be under 64 characters. In this case, the complete autogenerated index name, `jv_snapshots.$commitMetadata.properties.key_1_commitMetadata.properties.value_1`, exceeds that limit. This issue also occurs in MongoDB, if using an unusually long DB name.

Perhaps `MongoSchemaManager.ensureSchema()` could explicitly set a shorter index name? 

Thanks!

bartoszwalacik:
ok, consider contributing a PR

jpars:
PR is opened, however i just learned that this index isn't compatible with DocumentDB anyway, as it does not yet support compound indexes with multiple array fields. Amazon's docs seem to imply that this feature is coming, so maybe it's not worth addressing this issue until support is added. I can close the PR and perhaps follow up later if you prefer.

bartoszwalacik:
PR is merged. The index on commitProperties is quite important but maybe we can start with DocumentDB experimental support, so without this index in the first place.

bartoszwalacik:
released in 5.2.4 as *experimental support*

PR #790, opened 2019-02-19
#788 Explicitly set name for MongoDB snapshot commit property index

jpars:
Index name has been added. I only explicitly set the name of the one compound index that resulted in a very long generated name in mongo. Let me know if I should set all of them. Also, I tested that this does not cause issues on Mongo DB instances where the schema was already initialized with the old index name; the existing index is not replaced and no error is thrown by the Mongo driver. This was tested on Mongo 3.6.2.


bartoszwalacik:
awesome, 
btw your commit is'n linked with your github account, if you link it, you will automatically become a Javers contributor

PR #825, opened 2019-04-08
Add more Change Types for properties and implement finalDiff

franktominc:
Hello,

We needed a little more than the default javers behavior for an internal use case. For it, we had to change the way it handled changes internally. This pull request aims to propose the addition of those changes to javers core.

Here is a summary of what changed:
- There is a specific type for the addition and the removal of a property. (changing the value of a property to null is different than removing it from the model altogether).
- There is a new method `finalDiff` that returns a `Diff` object with everything on the object graph as `ObjectRemoved`, including the root of the graph(I'm not sure if that behaviour was possible before).

Regards,
Frank


bartoszwalacik:
Hi @franktominc, thanks for your concern about Javers. But this is Javers core API, it shouldn't be developed without prior discussion.

First, I don't understand the idea behind this new method

```
public Diff finalDiff(final Object removedObject)
```

Does it produces `ObjectRemoved` for all passed objects? If so, wouldn't is be more straightforward to simply call this?

```
compare(oldVersion, null)
``` 

Second, do we really need new classes to model changing from null to not-null? Why not simply add a boolean method in existing `PropertyChange`?

franktominc:
Hi @bartoszwalacik, I agree we should discuss those points.

About the first concern, yes, it produces `ObjectRemoved` for all passed objects. if we try to call `compare(oldVersion, null)` as you mentioned, it throws an `IllegalArgumentException` saying the arguments should not be null.
If we change that behavior, so it would accept one null argument, I think the `initial` method could be replaced to a call to `compare(null, newVersion)` too.

I don't think we need new classes to model changing from null to not-null, `ValueChange` and `ReferenceChange` are enough for that. But there is a subtle difference between changing a value to/from null and adding or removing it to a class. For instance:
```java
@TypeName("A")
abstract class A{
}

@TypeName("A")
class B extends A {
    @Id
    private UUID id;
    private String foo;
}

@TypeName("A")
class C extends A {
    @Id
    private UUID id;
    private String bar;
}
```
If we try to compare an instance of `B` and `C` using the current implementation, it would just say `foo` or `bar` was changed to `null`(depending on which instance is the new one) when what really happened was that the property was removed from the model. It's different than comparing two instances of `B` and updating `foo` to `null`.

bartoszwalacik:
Agreed, we should allow for passing nulls to compare:

```
//final
compare(oldVersion, null)
```

```
//initial
compare(null, newVersion)
```

It's enough to change null to `emptyGraph()`. Then, `initial()` could be deprecated.

Concerning properties issue. Now it's clear for me. Agreed that adding a property to a class is a different situation than simply changing existing property from null to not-null. Maybe we need better names :).

Please resolve conflicts and please remove `finalDiff()` and add null support to `compare()`. We will merge this PR.

franktominc:
I will work on the new design, however I'll not be able to do it this week. Expect those changes next week

bartoszwalacik:
no hurry,
btw nulls can be already passed to compare(), see https://github.com/javers/javers/blob/master/javers-core/src/main/java/org/javers/core/diff/DiffFactory.java#L92
maybe we just have to document it in compare() javadoc

franktominc:
Moved to #830

PR #830, opened 2019-04-22
Add new change types for properties on subclassing

franktominc:


franktominc:
Moved from #825

franktominc:
There are some failing tests here that I'm not sure how to fix. It seems to be using `ListChangeAppender` to calculate the changes between a list and another class instance. It tries to convert the class instance to a list and fails with a `ClassCastException`

bartoszwalacik:
huge PR :) give me a while

franktominc:
No worries, take your time :)

bartoszwalacik:
there are two problems with this PR

* You didn't provided tests for the new feature, so we don't know how it works and if it works. 
* You have changed some existing tests which should not be affected by this change. For example, there is no reason to change `SnapshotDifferIntegrationTest.groovy` or `JaversNapCategoryTreeIntegrationTest.groovy`. The more tests you change, the higher risk that you break existing code of Javers users

franktominc:
Regarding your points:
- I'll write some tests for it, however I'm afraid my groovy isn't that sharp, I'll try my best.
- The problem with the tests you mentioned was that they were testing for class equality, and this PR indeed **broke some things**. Now both `ValueChange` and `ReferenceChange` are abstract classes and each one of them has a final class for the added, updated and removed change.

Before the change:

Imgur

After the change:

Imgur

I updated those tests to reflect this change.

PR #836, opened 2019-05-16
fix toString() bug

Starkom:
toString always shows discouraging "Field Map<String, String>..." instead of correct "Field Map<String, MyClass>..."


bartoszwalacik:
thanks @Starkom , this fix will be released in 5.5.0

PR #841, opened 2019-05-22
sql multi tenancy sequences quick fix

DarthDude:
Hello,

We use hibernate multi tenancy, and we have a problem with the sequence map.
The pull request is a small quick fix, and there is probably something better to do ...


DarthDude:
I'm looking at the problem. When I execute from my ide, everything is okay.

DarthDude:
getSchema fail with H2, and finaly, i use database url instead

DarthDude:
I forgot, I'll do a test tomorrow :)

bartoszwalacik:
I'm waiting :)

DarthDude:
Hi,
I'm working on it, adding the multiTenancy hibernate property in HibernateConfig test. But I have some problems creating the schema, and I don't want to make too many changes in the code (I use liquibase)

DarthDude:
Hi,
Finally, I added liquibase in test dependency to avoid modifying your code (except for a few getter).
I also think that this could serve as an example. 
I'm sorry if everything's not perfect, but I was discovering gradle and groovy at the same time (and that's cool).
To finish, there are two datasources. It's just for the example because on my side, I use vault.

DarthDude:
Hi,
have you seen the test ?

bartoszwalacik:
I'm not familiar with the tenant concept, so could you please explain what's going on here, and why liquibase has jumped in?

Please remove reformatting changes, it makes CR harder,

bartoszwalacik:
as I see there are 3 ways of implementing multi-tenancy in Hibernate: SCHEMA, DATABASE, DISCRIMINATOR. Which one you are trying to add to Javers or maybe all of them?

DarthDude:
I used the database mode: _properties.setProperty("hibernate.multiTenancy", "DATABASE");_ in HibernateConfig.

I think that only one test is enough for the database and schema mode, since the key in the map is the url of the db.
On the other hand, the discriminator mode should also work (although the url of the db is useless), but this mode is not recommended.

In any case, the ultimate modification for sequence management is generic, and I don't think we can say that we strongly link hibernate and javers  (except for the tests)

DarthDude:
_"Please remove reformatting changes, it makes CR harder"_
I'll do it on Monday!

_"why liquibase has jumped in?"_
I used liquibase to avoid modifying or adding code.
It is only used for test purposes, to create the schema in the primary and secondary source datasource. In any case, in my multi-tenant context, I am obliged to do so, otherwise, I only have javers tables in the primary datasource.

In any case, the subject of my request was a quick fix. I don't allow myself to do more, but if you want to offer complete multitenant, I can help.

bartoszwalacik:
Your PR looks like a quick and dirty hack which would probably work for you. But why not simply add the fully blown multi tenancy support in Javers, and make it available for all Javers users?

DarthDude:
My management will not allow me to spend 1 week (probably more) to advance this development (I have a lot of work to do elsewhere). So I can work evenings or weekends, but it will take longer. And if this fix doesn't suit you in the end, refuse my pr.

DarthDude:
I'm still going to try to propose something quickly

bartoszwalacik:
@DarthDude there is no hurry, this is OSS project :) 
So let's say, we want Javers to be compatible with Hibernate's multi-tanency in SCHEMA and DATABASE modes.

So what we have to do:
- Javers has to know if an app is running in multi-tanency mode. New Javers props? Ask Hibernate?
- tenant context must be included in Sequences cache (that's pretty much what you have already done on this PR)
- tenant context must be included in all SQL operations performed by `JaversSqlRepository`, so Schema Manager and JQL queries. That's missing
- good E2E test

Something else?

DarthDude:
Hi,
_Javers has to know if an app is running in multi-tanency mode. New Javers props? Ask Hibernate?_

In my application, I have a Map<String, DataSource> in a bean, the key being the tenantId. This property alone could be enough to say that javers is running in multitenant mode.  Warning :  this map can be updated with new database (or schema) at runtime.

_tenant context must be included in all SQL operations performed by JaversSqlRepository, so Schema Manager and JQL queries. That's missing_

This is true for the schema manager, and that's why I used liquibase for the tests. I have no idea how much time is needed for this part.
Concerning the jql queries, it works very well in my context without having to do anything since the connectionProvider provided extends AbstractDataSourceBasedMultiTenantConnectionProviderImpl (hibernate). Maybe this part doesn't have to be handled in javers so ?

bartoszwalacik:
You are right, looks like JQL queries would just work, assuming that Hibernate provides Connections to the right database/schema.

bartoszwalacik:
closing, because this work is abandoned by the contributor

Issue #855, opened 2019-06-28
Large Commit ID is persisted with different value than logged


andreiamariei:
When using RANDOM CommitIdGenerator, Javers logs different Commit IDs than what's being actually persisted in the database.

The issue seems to be the conversion to `double` in `CommitId.valueAsNumber()`.

Test (quickly written in spock):
```
def "large commit id test"() {
    expect:
        new CommitId(8804939115786087745, 0).valueAsNumber() == 8804939115786087745.00
}
```
results in:
```
Condition not satisfied:

new CommitId(8804939115786087745, 0).valueAsNumber() == 8804939115786087745.00
|                                    |               |
8804939115786087745.0                |               false
                                     8804939115786087424.00

```

bartoszwalacik:
Hi , we will be happy to accept your pull request with the fix

bartoszwalacik:
fixed in 5.6.1

PR #856, opened 2019-06-28
Fix large CommitIds not fitting double

andreiamariei:
Issue #855


andreiamariei:
Updated commit message to include the issue ID, as per guidelines, hope you're OK with this force-push

PR #863, opened 2019-07-17
javers#860 snapshot violation test case

14327319:
Dear @bartoszwalacik ,
I created a simple test case which demonstrates issue#860. This exception occurs only when mappingStyle=bean in my case and disappears in the case of mappingStyle=field. I think this is a bug.
Hope this helps.


bartoszwalacik:
Ok, thanks, seems like a good start. Maybe you can come up with a fix also? Meanwhile check our new sponsorship page https://javers.org/sponsorship

Issue #894, opened 2019-10-09
Provide better collision safety for SYNCHRONIZED_SEQUENCE


duro1:
Looking to provide better collision avoidance when using the sequence number on distributed machines.

With SYNCHRONIZED_SEQUENCE marked as not safe for distributed, RANDOM generator marked as deprecated, and CUSTOM only able to compare based on timestamp, it is difficult to ensure proper ordering of events with distributed systems if there is any time drift.

providing a secondary comparator criteria allows for reasonable expectations of both ordering and collision avoidance for a smaller number of distributed systems connected to the same JaVers repository. This is not a foolproof solution, but better than what is currently available

bartoszwalacik:
thanks, I will manage your contribution in the next week

bartoszwalacik:
Well, RANDOM generator isn't deprecated, it's recommended if you have more than one instance of your app writing to the same Javers repository.
Yet, if you prefer to stay with SYNCHRONIZED_SEQUENCE, you can get better sorting, which is as I understand the subject of this issue. Give me a while to check your PR.

bartoszwalacik:
Fixed in 5.8.2, @duro1 thanks for your PR contribution.

PR #895, opened 2019-10-09
Update comparator for Synchronized Sequence to be safer with distributed systems

duro1:
Provides some level of security over duplicated commits when comparing shadows
Without this additional criteria, shadows with the same commit id will get the commit metadata from the earliest commit, rather than the commit associated to the shadow object

https://github.com/javers/javers/issues/894


bartoszwalacik:
thanks @duro1

bartoszwalacik:
btw, your commits are not linked to your github account, if you link them, you will be automatically counted as javers contributor

PR #896, opened 2019-10-10
Problem deserializing EmbeddedId after reastart

itisha:


bartoszwalacik:
thanks for contributing the test case, next steps will be done on this branch
https://github.com/javers/javers/tree/itisha-problem-deserializing-embedded-id-by-type-name

Issue #708, opened 2018-08-28
Auto Configuration missing ObjectAccessHook parameter (yml)


duayres:
On doc there is a clear explanation about the javers core auto configuration (Spring framework integration):
"Each property in the Spring application.yml file has the corresponding with*() method in the JaversBuilder."
I've tried to write the method without the "with" in snake case on my yaml as follows:
``` 
javers:
    ...
    objectAccessHook: my.package.CustomHibernateUnproxy
```
But it did not work, it seems missing.

Javers version: 3.11.

----
Edit: Maybe a subsection to denote the transactional builder and then the parameter can be implemented:
```
javers:
    ...
    transactional:
        objectAccessHook: my.package.CustomHibernateUnproxy
```

bartoszwalacik:
fixed in 5.8.3

PR #904, opened 2019-10-24
https://github.com/javers/javers/issues/708 - fix for missing spring boot ObjectAccessHook property

akrystian:


javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSqlProperties.java
```diff
@@ -5,9 +5,11 @@
 
 @ConfigurationProperties(prefix = "javers")
 public class JaversSqlProperties extends JaversSpringProperties {
+    private static final String DEFAULT_OBJECT_ACCESS_HOOK = "org.javers.hibernate.integration.HibernateUnproxyObjectAccessHook";
```
bartoszwalacik: `private static final String DEFAULT_OBJECT_ACCESS_HOOK = HibernateUnproxyObjectAccessHook.class.getName();`

javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/sql/JaversSqlProperties.java
```diff
@@ -5,9 +5,11 @@
...
 
     private boolean sqlSchemaManagementEnabled = true;
     private String sqlSchema;
+    private String objectAccessHook = DEFAULT_OBJECT_ACCESS_HOOK;
```
bartoszwalacik: `ObjectAccessHook` is a core interface, it is used also in `JaversMongoAutoConfiguration`

bartoszwalacik: so this props should be added in `JaversSpringProperties` and available for both SQL and Mongo starters

Issue #921, opened 2019-12-06
Allow JaversAuditableDelete work with id(s), e.g. deleteById(id)


IlyaNerd:
Currently we need to provide the entire entity to mark it as deleted in javers when using JaversAuditableDelete, would be nice to be able provide only id(s) for that. Probably same way as its done in JaversSpringDataAuditableRepositoryAspect

`
@JaversAuditableDelete
void deleteById(id)
`

bartoszwalacik:
we are open for contributions

IlyaNerd:
i know, just wanted the approve :)

bartoszwalacik:
Released in 5.8.7, thanks for your contribution

PR #927, opened 2019-12-09
#921: Allow JaversAuditableDelete work with id

IlyaNerd:
I couldnt figure out how to get entity class except specifying it in annotation, let me know if i missed smth

### DIFF
--- a/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversCommitAdvice.java
+++ b/javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversCommitAdvice.java
@@ -1,8 +1,13 @@
 package org.javers.spring.auditable.aspect;
 
 import org.aspectj.lang.JoinPoint;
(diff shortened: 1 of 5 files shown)


javers-spring/src/main/java/org/javers/spring/auditable/aspect/JaversCommitAdvice.java
```diff
@@ -30,18 +35,29 @@ public void commitSaveMethodArguments(JoinPoint pjp) {
         }
     }
 
-    public void commitDeleteMethodArguments(JoinPoint pjp) {
-        for (Object arg : AspectUtil.collectArguments(pjp)) {
-            commitShallowDelete(arg);
+    public void commitDeleteMethodArguments(JoinPoint jp) {
```
bartoszwalacik: nice PR, but please fix these nested ifs (extract method)

IlyaNerd: done

PR #938, opened 2020-02-12
Test Cases For Custom Date Serialization

seamuslowry:
I am new to the open source world and this is the first time I'm writing any groovy. Please attribute any mistakes to ignorance rather than malice

### DIFF
--- a/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
+++ b/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
@@ -146,9 +146,9 @@ public class JsonConverterBuilder {
     }
 
     public JsonConverter build() {
-        registerJsonTypeAdapters(UtilTypeCoreAdapters.adapters());
-        registerJsonTypeAdapters(Java8TypeAdapters.adapters());
-        registerJsonTypeAdapter(new AtomicTypeAdapter(typeSafeValues));
+        registerBuiltInAdapters(Java8TypeAdapters.adapters());
+        registerBuiltInAdapters(UtilTypeCoreAdapters.adapters());
+        registerBuiltInAdapter(new AtomicTypeAdapter(typeSafeValues));
 
         if (prettyPrint){
             gsonBuilder.setPrettyPrinting();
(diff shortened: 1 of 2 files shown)


bartoszwalacik:
Good test case. Custom JSON serializers should have priority over built-in serializers. Since you have already found the place in javers-core where serializers are beeing registered, think about contributing a PR with the fix. But remember, custom serializers manage only how your data are stored in JaversRepository. Diff logic is customized in CustomComparators.

seamuslowry:
It may take a couple days as I'm a little pressed for time right now, but I think I could do that. Would it be appropriate to just keep committing to this pull request?

bartoszwalacik:
take your time,
yes you can commit here

seamuslowry:
What do you think of that as a solution? It would be tedious to write the test cases such that they test all the adapters in `Java8TypeAdapters` and `UtilTypeCoreAdapters` for the ability to override, but I'll do it if you feel it is necessary.

I tried to generalize it but a lot of the types don't have a constructor valid for something like `clazz.getDeclaredConstructor().newInstance()` so I couldn't do a full generalization.

As it is, at least one type from each of `Java8TypeAdapters` and `UtilTypeCoreAdapters` is tested to ensure it can be overridden and I threw in a test to override the adapter for `Atomic` as well.

seamuslowry:
It wasn't sitting well with me to leave cases untested, so I wrote tests to verify that the 17 built in adapters I identified and affected could be properly overridden. I tried hard to generalize the tests so there wouldn't need to be 17 new adapter classes for this test, but I couldn't get to a solution that both tested all 17 cases and was fully generalized.

bartoszwalacik:
I'm on holiday, will back to you after a few days

bartoszwalacik:
Indeed, these 17 new adapter classes in the test case are not nice

bartoszwalacik:
I have fixed this test using composition, see https://github.com/javers/javers/pull/942/files

seamuslowry:
Clever. I never thought of using the real adapters as delegates. I'll keep that pattern in my back pocket. Thanks!

bartoszwalacik:
👍 thanks for your contribution
this fix just released in 5.8.10 https://javers.org/release-notes

PR #952, opened 2020-03-30
fix #951 : use gradle feature variant for optional dep

### LINKED ISSUE #951: Remove guava strong dep on javers-core
Gradle optional is sadly not optional on maven.
https://blog.gradle.org/optional-dependencies

I propose to use gradle feature variant, to put gradle as maven optional

Dufgui:


Dufgui:
thanks !

PR #954, opened 2020-04-06
Dependencies update to use newer versions of Spring etc

infoShare:
Some of the dependencies needs to be updated to allow using with newer version of Spring/Spring boot.


bartoszwalacik:
thanks,
released in 5.8.13

Issue #962, opened 2020-04-22
Legacy MongoClient conflict when declared explicitly as Bean


nishkarsh:
I am working on a spring boot project with JaVers integrated along with `spring-boot-starter-data-mongodb`.

I had to configure `MongoClient` for some encryption settings for which I need to explicitly create it as a Bean. The newer version of mongo java driver (since version 3.7) recommends that the `MongoClient` is created using `MongoClients` class which creates an instance of `com.mongodb.client.MongoClient` ([source](https://mongodb.github.io/mongo-java-driver/3.7/javadoc/com/mongodb/client/MongoClients.html)). When this happens, the application produces an error as JaVers is expecting a legacy version of MongoClient.

```
Field mongoClient in org.javers.spring.boot.mongo.JaversMongoAutoConfiguration required a bean of type 'com.mongodb.MongoClient' that could not be found.

The injection point has the following annotations:
	- @org.springframework.beans.factory.annotation.Autowired(required=true)

The following candidates were found but could not be injected:
	- Bean method 'mongo' in 'MongoAutoConfiguration' not loaded because @ConditionalOnMissingBean (types: com.mongodb.MongoClient,com.mongodb.client.MongoClient; SearchStrategy: all) found beans of type 'com.mongodb.client.MongoClient' mongoClient


Action:

Consider revisiting the entries above or defining a bean of type 'com.mongodb.MongoClient' in your configuration.
```

This forces me to stay with the legacy MongoClient that doesn't have a nice builder to accept various settings.

bartoszwalacik:
Hi, consider contributing a PR with mongo client version bump

bartoszwalacik:
@nishkarsh unfortunately, `MongoAutoConfiguration` from Spring still produces the old `com.mongodb.MongoClient` which makes things messy

nishkarsh:
@bartoszwalacik You are right, springboot autoconfigure still provides the legacy MongoClient as of now. However, I checked the [commit here](https://github.com/spring-projects/spring-boot/commit/d2d6dbdc00a9be05a59f03430c780bf9884a694a#diff-159afa3664e0a6e0a28383438d6a2cfe) that seems to have the change in `MongoAutoConfiguration` to provide `com.mongodb.client.MongoClient` which is tagged for 3.0.0 release. Since the `3.0.0.RC2` is already out, I believe the stable would be out soon too. Moving to the newer version, however, would probably take even longer.

[...]

PR #964, opened 2020-04-23
#962 Replace references to legacy MongoClient with newer version.

nishkarsh:
- Replaced all the imports of `MongoClient` from `com.mongodb.MongoClient` to `com.mongodb.client.MongoClient`
- Using `MongoClients` from `com.mongodb.client` package to construct `MongoClient` which is now the recommended way.

### DIFF
--- a/javers-persistence-mongo/src/test/groovy/org/javers/repository/mongo/EmbeddedMongoE2ETest.groovy
+++ b/javers-persistence-mongo/src/test/groovy/org/javers/repository/mongo/EmbeddedMongoE2ETest.groovy
@@ -1,6 +1,6 @@
 package org.javers.repository.mongo
 
-import com.mongodb.MongoClient
+import com.mongodb.client.MongoClient
 import com.mongodb.client.MongoDatabase
 import spock.lang.Shared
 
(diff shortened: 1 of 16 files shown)


bartoszwalacik:
thanks, but looks like the build is failing, I will take a look.

nishkarsh:
Hey @bartoszwalacik! The build was passing on local, seems like some more tests run on a different profile, I will also spend some time on it and see if I can fix it.

bartoszwalacik:
`JaversMongoStarterIntegrationTest` fails on local for me. there are issues in Spring DI, probably as a result of changed bean's type 

```
Description:

Field mongoClient in org.javers.spring.boot.mongo.JaversMongoAutoConfiguration required a bean of type 'com.mongodb.client.MongoClient' that could not be found.

The injection point has the following annotations:
	- @org.springframework.beans.factory.annotation.Autowired(required=true)
```

nishkarsh:
Oh, this spec never executed when I ran `./gradlew build`. Running it explicitly from IntelliJ confirms that it fails. Apparently I have missed a couple of places where the legacy MongoClient is being constructed. I will soon verify and push the fixes.

nishkarsh:
I have made all the changes. I guess the problem is with [Embedded Mongo](https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo) since it still depends on the legacy Mongo. I added a `MongoClient` Bean as a test configuration since Spring, in this case, provided legacy `MongoClient`.

This, however, changes things that might require a major version bump since `MongoClientOptions` have been replaced with `MongoClientSettings` too along with the migration of legacy `MongoClient`.

Let me know if I could help in any other way! Thanks!

PR #980, opened 2020-06-10
#981 Adding @Order(0) to JaversSpringDataAuditableRepositoryAspect.

K10Codes:
Adding @order(0) to JaversSpringDataAuditableRepositoryAspect.
Why ?
This aspect contains advices that run for all save/delete operations in
the application but since there is no order present with it. It takes
the default order which is highest precedence as it is using after-
advices. So giving it order 0 would help if some other advice needed to
execute before this.
That's why explicitly giving it 0.

### DIFF
--- a/javers-spring/src/main/java/org/javers/spring/auditable/aspect/springdata/JaversSpringDataAuditableRepositoryAspect.java
+++ b/javers-spring/src/main/java/org/javers/spring/auditable/aspect/springdata/JaversSpringDataAuditableRepositoryAspect.java
@@ -7,6 +7,7 @@ import org.javers.core.Javers;
 import org.javers.spring.auditable.AuthorProvider;
 import org.javers.spring.auditable.CommitPropertiesProvider;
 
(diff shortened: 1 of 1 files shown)


javers-spring/src/main/java/org/javers/spring/auditable/aspect/springdata/JaversSpringDataAuditableRepositoryAspect.java
```diff
@@ -17,6 +18,7 @@
  * Calls {@link Javers#commitShallowDelete(String, Object, Map)} on arguments passed to delete() methods.
  */
 @Aspect
+@Order(0)
```
K10Codes, line 21: Adding @Order(0) so that any other advice on the same point cut expression can be evaluated before this.

K10Codes:
> What is What?



> What is What?

Its an error at my end. I missed entering the title while opening the PR and default text "What?" persisted. Should i create a new PR with updated title ?

Issue #999, opened 2020-07-08
findChanges: results are not the same when using byClass, byInstanceId and anyDomainObject


ptaillard:
I spot an issue on findChanges when I try to get same data using 3 different way: **byClass, byInstanceId and anyDomainObject**
The test below create only one document (Object ID = Bob), and get the last 3 changes using byClass, byInstanceId and anyDomainObject.
You can copy/paste the following test into the org.javers.core.examples.ChangeLogExample files and run it as it.
And you will see that **byClass, byInstanceId returns 3** results contrary to **anyDomainObject which return only 2** => wrong.
Thanks

I'm using the springboot library:
```
<dependency>
  <groupId>org.javers</groupId>
  <artifactId>javers-spring-boot-starter-mongo</artifactId>
  <version>5.9.4</version>
</dependency>
```

Test case to reproduce the issue:
```
 @Test
    public void issue() {
        // given:
        Javers javers = JaversBuilder.javers().build();
        Employee bob = new Employee("Bob", 9_000, "ScrumMaster");
        javers.commit("hr.manager", bob);

        // do some changes and commit
        bob.setPosition("Developer");
        bob.setSalary(11_000);
        javers.commit("hr.director", bob);

        bob.addSubordinates(new Employee("Trainee One"), new Employee("Trainee Two"));
        javers.commit("hr.manager", bob);

        bob.setSalary(455);
        bob.setPostalAddress(new Address("bath"));
        javers.commit("hr.director", bob);


        bob.setSalary(999);
        bob.setPrimaryAddress(new Address("london", "344"));
        javers.commit("hr.director", bob);

        // when:
        List<Change> changes = javers.findChanges(
                QueryBuilder.byInstanceId("Bob", Employee.class).limit(3).withNewObjectChanges().build());
        String changeLog = javers.processChangeList(changes, new SimpleTextChangeLog());

        // then:
        System.out.println(changeLog);

        changes = javers.findChanges(
                QueryBuilder.byClass(Employee.class).withNewObjectChanges().limit(3).build());
        changeLog = javers.processChangeList(changes, new SimpleTextChangeLog());

        // then:
        System.out.println(changeLog);

        changes = javers.findChanges(
                QueryBuilder.anyDomainObject().withNewObjectChanges().limit(3).build());
        changeLog = javers.processChangeList(changes, new SimpleTextChangeLog());

        // then:
        System.out.println(changeLog);
    }
```

bartoszwalacik:
[...]

PR #1003, opened 2020-07-09
#999 findChanges: results different on byClass, byInstanceId AND anyD…

ptaillard:
…omainObject

https://github.com/javers/javers/issues/999


bartoszwalacik:
see https://github.com/javers/javers/issues/999#issuecomment-716106659

Issue #940, opened 2020-02-21
Support exclusion strategy for classes to be skipped from serialization


gpor0:
Javers supports registerIgnoredClass API to ignore specific class. It would be very useful in some cases to support ignoreStrategy / exclusionStrategy where implementor can define custom rule like ignoring by package or similar.

Gson already supports exclusion strategy but as far as i am aware of Gson configuration in Javers cannot be accessed at the moment.

bartoszwalacik:
Makes sense, we will accept your PR if you like to contribute.

API sketch :

```
javersBuilder.registerIgnoredClassesStrategy(IgnoredClassesStrategy strategy)
```

```
interface IgnoredClassesStrategy {

    boolen isIgnored(Class<?> domainClass);

}
```

bartoszwalacik:
```java
Javers javers = JaversBuilder.javers()
        .registerIgnoredClassesStrategy(c -> c.getName().startsWith("com.ignore.me"))
        .build(); 
```

bartoszwalacik:
released in 5.11.1

PR #1005, opened 2020-07-11
#940 support for exclusion strategy for classes

arkadiush:


bartoszwalacik:
Sorry but your implementation is wrong, ignored classes in Javers are those classes which are ignored by the javers diff & persistence engine.

PR #1006, opened 2020-07-11
Fixes #941

### LINKED ISSUE #941: Configurable table names (SQL)
Currently there is no way to change the default table names. 

- jv_global_id
- jv_commit
- jv_commit_property
- jv_snapshot

Providing a way of customizing these would be very nice.

shirishpandharikar:
Changes for supporting custom SQL table names

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
@@ -23,7 +23,7 @@ public class FixedSchemaFactory extends SchemaNameAware {
     public static final String GLOBAL_ID_FRAGMENT =   "fragment";     //since 1.2
     public static final String GLOBAL_ID_TYPE_NAME =  "type_name";    //since 2.0
     public static final String GLOBAL_ID_OWNER_ID_FK ="owner_id_fk";  //since 1.2
-    public static final String GLOBAL_ID_PK_SEQ =     "jv_global_id_pk_seq";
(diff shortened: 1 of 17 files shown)


javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
```diff
@@ -23,15 +23,15 @@
...
-    public static final String GLOBAL_ID_PK_SEQ =     "jv_global_id_pk_seq";
+    public static final String GLOBAL_ID_PK_SEQ =     "global_id_pk_seq";
 
     public static final String COMMIT_TABLE_NAME =    "jv_commit";
     public static final String COMMIT_PK =            "commit_pk";
     public static final String COMMIT_AUTHOR =        "author";
     public static final String COMMIT_COMMIT_DATE =   "commit_date";
     public static final String COMMIT_COMMIT_DATE_INSTANT =   "commit_date_instant";
     public static final String COMMIT_COMMIT_ID =     "commit_id";
-    public static final String COMMIT_PK_SEQ =        "jv_commit_pk_seq";
```
bartoszwalacik, line 34: As far as I see you have changed existing name of this seq from jv_commit_pk_seq to jv_commit_commit_pk_seq. You can't change existing names, it would break users' code.

shirishpandharikar: Thanks for the feedback @bartoszwalacik Indeed this will break user code if someone simply upgrades to a new version. It completely slipped out of my mind.

PR #1008, opened 2020-07-15
Shirishpandharikar issue 941

bartoszwalacik:
continuation for https://github.com/javers/javers/pull/1006
(https://github.com/javers/javers/issues/941)


javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/FixedSchemaFactory.java
```diff
@@ -17,34 +17,30 @@
...
     public static final String COMMIT_PROPERTY_NAME =       "property_name";
     public static final String COMMIT_PROPERTY_VALUE =      "property_value";
 
-    public static final String SNAPSHOT_TABLE_NAME =   "jv_snapshot";
     public static final String SNAPSHOT_PK =           "snapshot_pk";
     public static final String SNAPSHOT_COMMIT_FK =    "commit_fk";
     public static final String SNAPSHOT_GLOBAL_ID_FK = "global_id_fk";
     public static final String SNAPSHOT_TYPE =         "type";
     public static final String SNAPSHOT_VERSION =      "version";
-    public static final String SNAPSHOT_TABLE_PK_SEQ = "jv_snapshot_pk_seq";
```
bartoszwalacik, line 47: As far as I see you have changed existing name of this seq from `jv_commit_pk_seq` to `jv_commit_commit_pk_seq`. You can't change existing names, it would break users' code.

bartoszwalacik: maybe we should leave names of primary keys, sequences and indexes as is

PR #1021, opened 2020-09-18
Query snapshots from utc datetime

nickshoe:


bartoszwalacik:
Hi, well done, but tests are missing :). Please add one test similar to this onehttps://github.com/javers/javers/blob/master/javers-core/src/test/groovy/org/javers/core/JaversRepositoryE2ETest.groovy#L687
It will be executed on all repositories.

nickshoe:
@bartoszwalacik I've implemented the test but the build is failing..

I guess it is something related to the snapshot commit date initialization.

I'm not understanding why this occours (`commitDate` should be +2 hours since my system is Europe/Rome):

image

bartoszwalacik:
since fake clock is used here, now is exactly what you set in `setNow()`

bartoszwalacik:
this toTimestamp() conversion might be wrong, old java.util.Date API :(

Screenshot 2020-09-26 at 19 50 18

nickshoe:
@bartoszwalacik now I'm initializing the `dateProvider,now` with a system default zone offset datetime. I'm building the query dates and the expected dates using a system default zoned datetime and converting them to UTC. Let's see if it works.
(I think it's rather a trick.. is the codebase mainly intended to work with local datetimes?)

nickshoe:
@bartoszwalacik forget it. The build failed again.
I'm going to take a deep dive into the test code and into the repository code, but I'll do this in the next days.

bartoszwalacik:
@nickshoe I sort of fixed it here 
https://github.com/javers/javers/pull/1022
https://github.com/javers/javers/pull/1022/files#diff-6efcb86b5afdcabbc7d9a2adaf849a8fR138

In SQL, commitDateInstant is not Stored in the DATE col but in String col.
Don't remember why, probably it's about precision to nanos. Fortunately, ISO date format as String has the chronological ordering :)

bartoszwalacik:
merged in this PR https://github.com/javers/javers/pull/1022

bartoszwalacik:
released in 5.12
@nickshoe thanks for this high quality contribution

nickshoe:
@bartoszwalacik thank you for the support and guidance!

PR #1025, opened 2020-10-15
Migrate to Spock 2.0 and Groovy 3.0

### LINKED ISSUE #1024: Bump Spock version to 2.0-M2
This would give us Groovy 3.0

wpanas:
Fix #1024


javers-core/src/main/java/org/javers/common/reflection/JaversGetterFactory.java
```diff
@@ -42,6 +42,7 @@ private void findAllGetters(Class currentGetterSource) {
                     .filter(method -> isGetter(method) && !method.isBridge())
                     .filter(method -> !isOverridden(method, getters))
                     .map(getter -> createJaversGetter(getter, context))
+                    .filter(getter -> excludeDuplicatedProperties(getter, getters))
```
bartoszwalacik, line 45: are we sure that filter() for second getter will be executed after getters::add for first getter?

wpanas: `isOverriden` works the same way

bartoszwalacik: right :)

Issue #576, opened 2017-08-01
Abandoned connections after schema validation


dbacinski:
During app startup Javers is validating if database schema is correct.

When i have enable logging of abandoned connections I am getting a stack trace from Javers, like this:
```
WARNING: Connection has been abandoned PooledConnection[org.postgresql.jdbc.PgConnection@56e64c5]:java.lang.Exception
	at org.apache.tomcat.jdbc.pool.ConnectionPool.getThreadDump(ConnectionPool.java:1102)
	at org.apache.tomcat.jdbc.pool.ConnectionPool.borrowConnection(ConnectionPool.java:807)
	at org.apache.tomcat.jdbc.pool.ConnectionPool.borrowConnection(ConnectionPool.java:651)
	at org.apache.tomcat.jdbc.pool.ConnectionPool.getConnection(ConnectionPool.java:198)
	at org.apache.tomcat.jdbc.pool.DataSourceProxy.getConnection(DataSourceProxy.java:132)
	at org.springframework.jdbc.datasource.DataSourceUtils.fetchConnection(DataSourceUtils.java:151)
	at org.springframework.jdbc.datasource.DataSourceUtils.doGetConnection(DataSourceUtils.java:115)
	at org.springframework.jdbc.datasource.DataSourceUtils.getConnection(DataSourceUtils.java:78)
	at info.stsa.maint.config.JaversConnectionProvider.getConnection(JaversConnectionProvider.kt:10)
	at org.javers.repository.sql.SqlRepositoryBuilder$1.getConnection(SqlRepositoryBuilder.java:77)
	at org.polyjdbc.core.transaction.ExternalTransactionManager.openTransaction(ExternalTransactionManager.java:18)
	at org.polyjdbc.core.schema.SchemaManagerFactory.createManager(SchemaManagerFactory.java:44)
	at org.polyjdbc.core.DefaultPolyJDBC.schemaManager(DefaultPolyJDBC.java:74)
	at org.javers.repository.sql.schema.JaversSchemaManager.ensureSchema(JaversSchemaManager.java:39)
	at org.javers.repository.sql.JaversSqlRepository.ensureSchema(JaversSqlRepository.java:88)
	at org.javers.core.JaversBuilder.build(JaversBuilder.java:113)
	at info.stsa.maint.config.MaintenanceDataSourceConfiguration.maintJavers(MaintenanceDataSourceConfiguration.kt:126)
```
This is how connection provider is implemented, I am using Spring Boot 2.0.0.M2:
[...]

PR #1046, opened 2020-12-24
Abandoned connections in schema manager

ivanjermakov:
This method was causing a connection leak due to opened connection wasn't closed.
This was mentioned in #576.


bartoszwalacik:
Connections are managed by an application, typically in a connection pool. They should not be closed by Javers

Issue #939, opened 2020-02-19
Registering Type Adapters with Spring Boot Starter


danesp2012:
Is there a recommended way to register TypeAdapters or custom comparators when using the Spring Boot starter?

bartoszwalacik:
There is no an easy way. Possible solution would allow user to construct `JaversBuilder` bean (with all custom configuration). 

Looks like an easy task to implement using `@ConditionalOnMissingBean`, take a look at starter class - `JaversSpringMongoApplicationConfig`. We will accept your PR if you want to contibute

bartoszwalacik:
released in 5.15
see https://javers.org/release-notes

PR #1061, opened 2021-03-03
#939 allow customizing javers builder

IlyaNerd:
#939


bartoszwalacik:
Hi, thanks for a contribution but please add tests

IlyaNerd:
frankly, not sure what to test here..
that custom adapter can be picked up? - not sure how to test it, as there is no explicit result to verify, somehow extract the ser/deser from gson?
plugin class - that it executes foreach?

bartoszwalacik:
You can test custom typeAdapters like in `JsonConverterCustomTypeAdapterTest` but without register them using this new feature. You can write your test in both starters close to `JaversMongoStarterIntegrationTest` and `JaversSqlStarterIntegrationTest`

PR #1064, opened 2021-03-04
Enable Sql Column Change

gustavodsf:
In a project of CESAR the client has a specific rule of the database creation, so we needed to customizes the names of the entire database, so we needed to add this type of configuration to javers. So your have to add the configuration below on your application.properties


javers-persistence-sql/README.md
```diff
@@ -0,0 +1,44 @@
+### This repo is a customization of javers
```
bartoszwalacik, line 1: please remove this story about Cesar ...

javers-persistence-sql/src/main/java/org/javers/repository/sql/SqlRepositoryBuilder.java
```diff
@@ -103,6 +130,186 @@ public SqlRepositoryBuilder withCommitPropertyTableName(String commitPropertyTab
         }
         return this;
     }
+    
+    public SqlRepositoryBuilder withGlobalIdPKColumnName(String globalIdPKColumnName) {
```
bartoszwalacik, line 134: sorry but this amount of boilerplate code is unacceptable.

All this names should be modeled as enum:

```
enum DbNames {
globalIdPKColumnName, globalIdLocalIdColumnName, ...
}
```

an then, the Builder should hold a map with concrete  values for all names `Map<DbNames,String>`

then, we don't need all these methods in the Buidler. Two methods should be enough:

```
public SqlRepositoryBuilder withDbName(DbNames dbName, String value)

public SqlRepositoryBuilder withDbNames(Map<DbNames,String)
```

bartoszwalacik (changes requested):

bartoszwalacik:
I have serious doubts about this feature. You can already change the names of javers tables.

Why on earth change the col names? Could you please give some example? How would you change the col names in javers table and why?

gustavodsf:
OK, but the project is for a big oil company, and it has severe database policy, all system has to follow his database column a table name pattern, if don't the system could not enter in production. So if i don't put to follow the pattern i have to create my own, but if this modification is not good for javers, we can give up so you dont waste your time.

bartoszwalacik:
If this big oil company will pay me via https://github.com/sponsors/bartoszwalacik I will do this feature

gustavodsf:
I'll talk to them, but let's make it clear if I make these changes, will you still only merge this into the javers repository if they pay as sponsors? Or you never merge this into javers repository ? with this i don't waste your and my time, and if necessary i start to find another solution.  Thanks

bartoszwalacik:
This new feature (not exactly this code on PR but close) won't be added as a regular feature, beacuse I think it's a corner case. But the feature can be added if sponsored. Sponsored corner cases are fine for me.

Issue #1009, opened 2020-07-16
Spring @Transactional update support


KarolKlimonczyk:
I try to use Javers in my application, and I noticed, that Javers doesn't audit objects which are updated in a transaction. 
E.g. when I use @Transactional annotation above my method, I don't have to save an object manually, it will be updated because it is in a managed state, but Javers doesn't audit changes like this. 

Please take a look at the example below:

```
@Service
@RequiredArgsConstructor
public class UserService {
    private final UserRepository userRepository;

    public User updateUser(Long id, String name) {
        User user = this.userRepository.findById(id).get();
        user.setName(name);
        return this.userRepository.save(user);
    }

    @Transactional
    public User transactionalUpdate(Long id, String name) {
        User user = this.userRepository.findById(id).get();
        user.setName(name);
        return user;
    }
}
```

The first method works like charm, but the second doesn't work at all. 

As far as I know, it can't work because Javers internally uses aspect around the 'save' method from CrudRepository, as follows: 
```
    @AfterReturning(value = "execution(public * save(..)) && this(org.springframework.data.repository.CrudRepository)", returning = "responseEntity")
    public void onSaveExecuted(JoinPoint pjp, Object responseEntity) {
        onSave(pjp, responseEntity);
    }

```

I use the newest 'javers-spring-boot-starter-sql'. 

Is there any chance to support audit when an entity is managed by EntityManager and then committed in a transaction? 
Or is it already supported, but I missed something? If so, sorry to raise a ticket. 
Thanks in advance.

bartoszwalacik:
Hi, you are right, Javers doesn't track changes when an object is saved implicitly by the EntityManager. We are open for contributions

KarolKlimonczyk:
Thanks for your quick reply. I'll try to play with that in my spare time.

edudant:
Hi, we have the same problem (as would anyone using JPA / transactions). 

The implementation looks pretty straightforward to me - just add a JPA EntityListener with @PostPersist/Update and @PostDelete and call javers.commit or javers.commitShallowDelete (some magic to get author and commit properties might be necessary).

Is the right way to implement it or do I miss something?

bartoszwalacik:
Yes, it shouldn't be hard, if you contribute a PR we will merge

bullerhenrik:
I'm curious about the status of this issue, since EntityManager integration is kind of a deal breaker for my use case. I realize #1009 is an old PR, but is it going to be updated/merged? or is someone else working on this?

[...]

PR #1066, opened 2021-03-08
#1009 integration with JPA EntityManager - EntityListeners

IlyaNerd:
#1009

### DIFF
--- a/javers-spring-jpa/aop.xml
+++ b/javers-spring-jpa/aop.xml
@@ -1,0 +1,10 @@
+<!DOCTYPE aspectj PUBLIC "-//AspectJ//DTD//EN" "http://www.eclipse.org/aspectj/dtd/aspectj.dtd">
+<aspectj>
+    <weaver>
+        <include within="org.javers.spring.jpa.JaversEntityListener"/>
+    </weaver>
+
+    <aspects>
+        <aspect name="org.springframework.beans.factory.aspectj.AnnotationBeanConfigurerAspect"/>
+    </aspects>
+</aspectj>
--- a/javers-spring-jpa/src/main/java/org/javers/spring/jpa/JaversEntityListener.java
+++ b/javers-spring-jpa/src/main/java/org/javers/spring/jpa/JaversEntityListener.java
@@ -1,0 +1,130 @@
+package org.javers.spring.jpa;
+
+import org.javers.common.collections.Maps;
+import org.javers.core.Javers;
+import org.javers.spring.auditable.AuthorProvider;
+import org.javers.spring.auditable.CommitPropertiesProvider;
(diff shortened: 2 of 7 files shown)


bartoszwalacik:
great, I need a while for code review

javers-spring-jpa/aop.xml
```diff
@@ -0,0 +1,10 @@
+<!DOCTYPE aspectj PUBLIC "-//AspectJ//DTD//EN" "http://www.eclipse.org/aspectj/dtd/aspectj.dtd">
```
bartoszwalacik, line 1: Can we do this without xml? I mean in Java Config

javers-spring-jpa/src/main/java/org/javers/spring/jpa/JaversEntityListener.java
```diff
@@ -0,0 +1,130 @@
...
+import javax.persistence.PostRemove;
+import javax.persistence.PostUpdate;
+import java.util.Map;
+
+import static java.util.Objects.requireNonNull;
+
+/**
+ * JPA entity listener to commit auditing changes. To get this one flying be
+ * sure you add @EnableJaversEntityListeners to your application config
+ * and add this entity listener in your {@code orm.xml} as follows:
```
bartoszwalacik, line 21: again, xml is passé, is there another option to kick it of without xml?

IlyaNerd: thats just a documentation how user can trigger this listener, if they like/need xml there is an option, which hibernate provides, not javers
and there are two options specified - xml or @EntityListeners, whichever user need
mb we should clarify this piece of javadoc if its confusing..

javers-spring-jpa/aop.xml
IlyaNerd: so there is no such thing as java config at this point, its instruction for aspectj compiler (ajc) what to weave at compile time.
the link you gave has one diff with our case - we need to inject 3 beans somehow into a listener which is not created by spring, but by a EntityManager at startup. If you look at this class in build directory, you will see that in default constructor there is additional code generated by aspectJ at compile time. For that we are using spring feature - @Configured with aspectJ - https://www.baeldung.com/spring-inject-bean-into-unmanaged-objects . Same thing does spring-data-jpa with their AuditingEntityListener.
There is also another option to autowire into such objects - special class with ApplicationContext static field or having this listener with static fields and resolving them later, which in my opinion is kinda ugly. - https://stackoverflow.com/a/60651528
We could remove the xml, but then ajc will weave entire module and will find 3 classes to weave (this one, transactional decorator and javers spring data aspect), not sure if its desirable to alter their logic.

Issue #1055, opened 2021-01-27
Invalid Json fetched from jv_snapshot.state column oracle sql database


WojtekPWas:
**Clear description of my expectations versus reality**

We have got recently an error 
com.google.gson.JsonSyntaxException: com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 3 path $
	at com.google.gson.Gson.assertFullConsumption(Gson.java:864) ~[gson-2.8.2.jar!/:na]
	at com.google.gson.Gson.fromJson(Gson.java:854) ~[gson-2.8.2.jar!/:na]
	at com.google.gson.Gson.fromJson(Gson.java:802) ~[gson-2.8.2.jar!/:na]
	at com.google.gson.Gson.fromJson(Gson.java:774) ~[gson-2.8.2.jar!/:na]
	at org.javers.core.json.JsonConverter.fromJsonToJsonElement(JsonConverter.java:70) ~[javers-core-3.10.2.jar!/:na]

The reason was in line 22 in org/javers/repository/sql/finders/CdoSnapshotMapper.java (javers-persistence-sql module)
             .withSnapshotState(resultSet.getString(SNAPSHOT_STATE))
The method withSnapshotState in CdoSnapshotSerialized class gets string as a parameter. This parameter has a value, which in debugger looks very similar to 
https://stackoverflow.com/questions/46198533/java-sql-resultset-getstring-returns-string-with-extra-characters

Contents of column state (of type CLOB) in the table jv_snapshots in the Database (retrieved with SQL Developer tool) looks ok.

Method getString is not the right way to retrieve data from the CLOB column, although it brings correct results in most cases.

The solution was to use getClob method as described in https://stackoverflow.com/questions/19486648/how-to-retrive-the-clob-value-from-oracle-using-java/19486749.

I prepared custom Javers version basen on 3.10.2 and the problem disappeared.

I looked at the current version of Javers 5.14.0.  getString(SNAPSHOT_STATE) is still there.
That is why I suggest the issue. Of course I would like to contribute and implement it in the current javers version.

**Steps To Reproduce**
I have no idea how to produce test case. Problem arise with fetching data from the database. The data is correct. STATE column of problematic row contains many characters, but there were cases with more characters and no error. I suppose the problem has to do with how Oracle Server maintains the data, that data not necessarily occupy continuous space.

**Javers' Version**
3.10.2
[...]

PR #1067, opened 2021-03-09
Issue #1055 Invalid Json fetched from jv_snapshot.state column oracle…

WojtekPWas:
… sql database.
Two small changes and one test class.
I could not reproduce the problem.
Test returns correct data also without my modification.
This is probably because of database maintenance

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java
@@ -183,7 +183,7 @@ class SnapshotQuery {
                     .withCommitId(resultSet.getBigDecimal(COMMIT_COMMIT_ID))
                     .withCommitPk(resultSet.getLong(COMMIT_PK))
                     .withVersion(resultSet.getLong(SNAPSHOT_VERSION))
-                    .withSnapshotState(resultSet.getString(SNAPSHOT_STATE))
+                    .withSnapshotState(resultSet.getClob(SNAPSHOT_STATE))
                     .withChangedProperties(resultSet.getString(SNAPSHOT_CHANGED))
                     .withSnapshotType(resultSet.getString(SNAPSHOT_TYPE))
                     .withGlobalIdFragment(resultSet.getString(GLOBAL_ID_FRAGMENT))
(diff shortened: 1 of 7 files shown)


bartoszwalacik (changes requested):
needs work, see my comments in https://github.com/javers/javers/issues/1055

javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java
```diff
@@ -9,24 +9,53 @@
...
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
-import static org.javers.repository.sql.schema.FixedSchemaFactory.*;
-import static org.javers.repository.sql.session.Parameter.*;
+import static org.javers.repository.sql.schema.FixedSchemaFactory.COMMIT_AUTHOR;
```
bartoszwalacik, line 23: no :)

javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java
```diff
@@ -9,24 +9,53 @@
...
+import static org.javers.repository.sql.schema.FixedSchemaFactory.SNAPSHOT_STATE;
+import static org.javers.repository.sql.schema.FixedSchemaFactory.SNAPSHOT_TYPE;
+import static org.javers.repository.sql.schema.FixedSchemaFactory.SNAPSHOT_VERSION;
+import static org.javers.repository.sql.session.Parameter.bigDecimalParam;
+import static org.javers.repository.sql.session.Parameter.instantParam;
+import static org.javers.repository.sql.session.Parameter.localDateTimeParam;
+import static org.javers.repository.sql.session.Parameter.longParam;
+import static org.javers.repository.sql.session.Parameter.stringParam;
 
 class SnapshotQuery {
```
bartoszwalacik, line 50: If Dialect is needed in SnapshotQuery, maybe you can get it from selectBuilder.session.dialect?

bartoszwalacik (changes requested):

javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java
WojtekPWas: I have pushed a new version. I have used session.dialect, but the classes Session, Dialect and Dialects are more public now and it is of course a drawback. I can not say whether it is better comparing with the previous version with additional injections.

bartoszwalacik: ok, I will take a look

Issue #1157, opened 2021-11-02
Comparison of Set of Set non-deterministic


samba2:
I have a data structure which contains a set of set. When comparing with javers I sometimes get a diff, sometimes not.  The test data is static so I would expect an identical output. 

**Steps To Reproduce**

When I run this class containing the two identical objects `a1` and `a2`:

```java
import lombok.Builder;
import lombok.Data;
import org.javers.core.Javers;
import org.javers.core.JaversBuilder;

import java.util.Set;

public class MyTest {

    @Data
    @Builder
    public static class A {
        private Set<B> bs;
    }

    @Data
    @Builder
    public static class B {
        private Set<C> cs;
    }

    @Data
    @Builder
    public static class C {
        private String foo;
    }

    public static void main(String[] args) {
        Javers javers = JaversBuilder.javers().build();

        A a1 = A.builder().bs(Set.of(
                        B.builder()
                            .cs(
                                Set.of(
                                    C.builder().foo("a").build(),
                                    C.builder().foo("b").build()
                                        )).build())).build();

        A a2 = A.builder().bs(Set.of(
                B.builder()
                        .cs(
                                Set.of(
                                        C.builder().foo("a").build(),
                                        C.builder().foo("b").build()
                                )).build())).build();


        for (int i = 0; i < 10; i++) {
            var diff = javers.compare(a1, a2);
            if (diff.hasChanges()) {
                System.out.printf("\n===== Diff at round %d ======\n", i);
                System.out.println(diff);
            }
        }
    }
}
```
... I receive the following output:
```
20:51:47.647 [main] DEBUG org.javers.core.JaversBuilder - starting up JaVers ...
20:51:47.724 [main] INFO org.javers.core.JaversBuilder - mappingStyle: FIELD
20:51:47.732 [main] INFO org.javers.core.JaversBuilder - loading JodaAddOns ...
20:51:47.786 [main] INFO org.javers.core.JaversBuilder - loading GuavaAddOns ...
20:51:47.841 [main] DEBUG org.javers.TypeMapper - javersType of 'LocalDateTime' mapped explicitly to ValueType
20:51:47.841 [main] DEBUG org.javers.TypeMapper - javersType of 'LocalDate' mapped explicitly to ValueType
20:51:47.842 [main] INFO org.javers.core.JaversBuilder - using fake InMemoryRepository, register actual Repository implementation via JaversBuilder.registerJaversRepository()
[...]

PR #1158, opened 2021-11-03
#1157 Comparing Set of Set is non-deterministic

samba2:
On my machine this test fails with the following output:

```
...
  detectedDiffs == 0
  |             |
  48            false
...
```
The number of detected diffs is changing with each test invocation.


samba2:
changed to "draft" to prevent unintentional merge

bartoszwalacik:
that's strange. I will take a look

samba2:
@bartoszwalacik thank you for your effort !

Issue #1075, opened 2021-03-13
Allow CustomValueComparators to handle nulls


bartoszwalacik:
Currently, before calling a CustomValueComparator impl, Javers checks nulls and assumes that :

```
null != any non-null value
```

This assumption simplifies an impl, but from the other hand it 
makes impossible to write a Comparator which treats nulls as empty values.
See https://stackoverflow.com/questions/66497652/javers-comparison-string-collection-and-boolean-with-defined-rules

```java
public interface CustomValueComparator<T> {
    /**
     * Called by JaVers to compare two Values.
     *
     * @param a not null
     * @param b not null
     */
    boolean equals(T a, T b);
```

**New feature**
Skip the Javers internal null-safety logic and allow passing nulls to CustomValueComparators. 

Unfortunately it could break existing users' code

**Test case**
[CaseWithCustomComparatorBlejwas.groovy](https://github.com/javers/javers/commit/045d751f4b7a8cdb959b246bc9b85538993befc2)

djavaphp:
Hi 
do you have any update on this? 
I want to ignore upcoming fields which is empty or null. 
Is there any way to do?

bartoszwalacik:
Hi, if you want to speed up things, feel free to contribute a PR

ratskinmahoney:
@bartoszwalacik - I'd be happy to raise a PR for this, but wanted to check if you had any thoughts on approach before I do.

Currently, what I am thinking is that I would modify `CustomValueComparator` to include a method to indicate whether wrapping with the `CustomValueComparatorNullSafe` should be skipped (with a default that returns false):

```java
public interface CustomValueComparator<T> {
  
  // ...

  default boolean skipNullSafety() { return false; }
}
```

This could then be extended with an interface that clearly describes it's behaviour:

```java
public interface NonNullSafeCustomValueComparator<T> extends CustomValueComparator<T> {

  @Override
  boolean skipNullSafety() { return true; }
}
```

Then `PrimitiveOrValueType` can be modified to check this flag:

```java
  PrimitiveOrValueType(Type baseJavaType, CustomValueComparator<T> comparator) {
    super(baseJavaType);
    this.valueComparator = comparator == null || comparator.skipNullSafety() ? comparator : new CustomValueComparatorNullSafe<>(comparator);
  }
```

That seems simple enough, as well as backwards compatible. I'm not sure if the new comparator interface should be advertised more clearly (e.g. through a new method on the builder, but that seems a bit bloating).

[...]

PR #1159, opened 2021-11-04
#1075 - Allow CustomValueComparators to handle nulls

ratskinmahoney:
#1075 - [This comment](https://github.com/javers/javers/issues/1075#issuecomment-959192203) begins the outline of the approach.


ratskinmahoney:
#1075

Issue #1208, opened 2022-07-05
WriteConflict when using Mongo transactions


jkloecknerneuland:
When using Javers with MongoDB and transaction support enabled, writing multiple commits at the same yields a WriteConflict error. An expected behavior would probably be to retry the commit or transaction.

**Steps To Reproduce**
I have a **runnable test case** which isolates the bug and allows Javers Core Team to easily reproduce it. I have pushed this test case to my fork of this repository:
https://github.com/jkloecknerneuland/javers/blob/master/javers-spring-boot-starter-mongo/src/test/groovy/org/javers/spring/boot/mongo/JaversMultipleTransactionsMongoTest.groovy

**Javers' Version**
6.6.5

**Additional context**
This is probably because the counter in the jv_head_id collection cannot be updated, because its blocked by another transaction.

Zakhar2093:
I got the same bug. It was really difficult to find the cause.  I have just the usual Controller which only inserts one document into the collection. When the backend performs several simultaneous requests I got this error. The solution was to delete `javers-spring-boot-starter-mongo` from pom.xml

jkloecknerneuland:
@Zakhar2093 Thanks for the suggestion, but unfortunately removing the `javers-spring-boot-starter-mongo` dependency is not an option for me, as I need the provided functionality of it.

jkloecknerneuland:
Sadly, this problem prevents me from using Javers, but I would like to contribute to the project to fix this.
Is there anything I can do?

kulasekp:
Hi, 
@bartoszwalacik @jkloecknerneuland 
We have faced exactly the same issue: Mongo transaction ends with WriteConflict because it fails to update jv_head_id:
```
	at org.javers.repository.mongo.MongoRepository.transactionalUpdate(MongoRepository.java:413)
	at org.javers.repository.mongo.MongoRepository.persistHeadId(MongoRepository.java:256)
	at org.javers.repository.mongo.MongoRepository.persist(MongoRepository.java:99)
	at org.javers.spring.mongodb.TransactionalMongoRepository.lambda$persist$0(TransactionalMongoRepository.java:76)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:140)
	at org.javers.spring.mongodb.TransactionalMongoRepository.persist(TransactionalMongoRepository.java:74)
	at org.javers.repository.api.JaversExtendedRepository.persist(JaversExtendedRepository.java:154)
```

@jkloecknerneuland 
[...]

PR #1209, opened 2022-07-12
#1208 add retry and test case for concurrent mongo transactions

jkloecknerneuland:


bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1311

PR #1215, opened 2022-08-22
Bug when comparing nested objects

eternalconcert:
When having two objects of which one has a child object, JaVers is not able to recognize the difference,
    when the properties of the child object have the same values as the corresponding default value in
    org.javers.common.collections.Defaults
    The first test ends successful, because the child Item has the value property `1`, which differs from the default
    in Defaults (0).
    The second test fails, because the child Item has the value property `0`, which is the same as the default in
    Defaults (0).
    In both tests the itemParentA, which has a child Item is compared to the itemParentB, which has no child Item.
    Please ensure, that removed nested objects are removed properly. I think the Problem is the FakeNode which uses the
    defaults from Defaults in the methods getDehydratedPropertyValue and getPropertyValue.
    The behaviour can be reproduced with other data types.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/CoreConfiguration.java
+++ b/javers-core/src/main/java/org/javers/core/CoreConfiguration.java
@@ -23,11 +23,14 @@ public class CoreConfiguration {
 
     private final boolean terminalChanges;
 
+    private final boolean usePrimitiveDefaults;
+
     private final CommitIdGenerator commitIdGenerator;
 
     private final Supplier<CommitId> customCommitIdGenerator;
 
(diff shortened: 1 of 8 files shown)


bartoszwalacik:
Thanks for reporting the issue and writing the test case, can you come up with a fix on this PR?

eternalconcert:
> Thanks for reporting the [issue](https://github.com/javers/javers/issues/1216) and writing the test case, can you come up with a fix on this PR?

I will try my very best.

eternalconcert:
I updated my fork with a new config parameter, to address this issue. Please review and merge my pull request.

PR #1222, opened 2022-10-01
Fix validation in QueryParamsBuilder (fixes #1183)

### LINKED ISSUE #1183: Possible issue with skip value check
While reading the code, I found this snippet on`master`:
https://github.com/javers/javers/blob/f49b0aea4c371b90ada54144843e3e1eae81fe65/javers-core/src/main/java/org/javers/repository/api/QueryParamsBuilder.java#L111
```
public QueryParamsBuilder skip(int skip) {
        Validate.argumentCheck(limit >= 0, "Skip is not a non-negative number.");
        this.skip = skip;
        return this;
    }
```

Shouldn't it be `        Validate.argumentCheck(skip >= 0, "Skip is not a non-negative number.");` ?

marcoebbinghaus:
#1183


bartoszwalacik:
thanks, released in 6.7.1

Issue #1223, opened 2022-10-07
Partial and case-insensitive search for Author and Commit Property


krisztaa0404:
**Is your feature request related to a problem? Please describe.**
Currently there is no option to search for an author by only providing a part of the username. The correct result set will be only returned if the exact username is provided. Commit property offers partial search, but it's sensitive for lowercase and uppercase letters.

**Describe the solution you'd like**
I would like to implement partial and case insensitive search for both author and commit property value fields.
eg. `byAuthorLikeIgnoreCase`, `byCommitPropertyLikeIgnoreCase`

**Describe alternatives you've considered**
None

**Additional context**
 I would like to contribute by creating a pull request, let me know what you think about the request.

bartoszwalacik:
Hi @krisztaa0404 , that's a good idea, feel free to contribute a PR

krisztaa0404:
#1224 Pull request created for issue

bartoszwalacik:
released in 6.8.0, thanks @krisztaa0404  👍

PR #1224, opened 2022-10-09
#1223 Partial and case-insensitive search for Author and Commit Property

krisztaa0404:
- Created new queries to handle partial and case-insensitive search on author
- Changed commitPropertyLike query to also ignore case when searching
- Created unit tests

Issue #1223

@bartoszwalacik Please take a look at it, if you have time, thanks in advance :)

### DIFF
--- a/javers-core/src/main/java/org/javers/repository/jql/QueryBuilder.java
+++ b/javers-core/src/main/java/org/javers/repository/jql/QueryBuilder.java
@@ -494,11 +494,8 @@ public class QueryBuilder {
     }
 
     /**
-     * Only snapshots with a given commit property partially containing a given value.<br/>
-     * Equivalent to SQL clause: WHERE property_value LIKE '%value%'
-     * <br/><br/>
-     *
-     * The matching is case insensitive on MongoDB and on most SQL databases.
+     * Only snapshots with a given commit property partially containing a given value, ignoring case.<br/>
+     * Equivalent to SQL clause: WHERE lower(property_value) LIKE lower('%value%')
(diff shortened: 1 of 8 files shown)


javers-core/src/main/java/org/javers/repository/jql/QueryBuilder.java
```diff
@@ -509,6 +509,17 @@ public QueryBuilder withCommitPropertyLike(String name, String value){
         return this;
     }
 
+    /**
+     * Only snapshots with a given commit property partially containing a given value, ignoring case.<br/>
+     * Equivalent to SQL clause: WHERE lower(property_value) LIKE lower('%value%')
+     *
+     */
+    public QueryBuilder withCommitPropertyLikeIgnoreCase(String name, String value) {
```
bartoszwalacik, line 517: the javadoc for `withCommitPropertyLike()` says pretty much the same. I suggest to force the existing method to be case insensitive

krisztaa0404: Do you mean I should change the existing function to also ignore case when searching, rather than creating a new separate method? :)

bartoszwalacik: exactly

krisztaa0404:
I made the requested changes, let me know if this is what you had in mind.

bartoszwalacik:
@krisztavasas I have merged your change via my PR, so this PR can be safely close
thanks a lot for your high quality contribution, I will release it soon

PR #1248, opened 2022-12-29
Error while restarting the server using Postgres

PankajKhairnar07:
I have created the standalone spring boot application with the "Postgres" database. Whenever the server starts the first time all tables are getting appropriately created in the schema used by my application. When I restart the application getting an error as "ERROR: relation "jv_global_id" already exists". I checked the code "PgDatabaseMetaData" already skipping the system default namespace. so if we are getting results from the result set means we already have a table that exists in the Postgres database.
Note: Using "demo" as schema name but for others, it can be any random string.

### DIFF
--- a/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/JaversSchemaManager.java
+++ b/javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/JaversSchemaManager.java
@@ -254,6 +254,7 @@ public class JaversSchemaManager extends SchemaNameAware {
                 } while(tableSchemaName != null
                         && !tableSchemaName.equalsIgnoreCase("public")
                         && !tableSchemaName.equals("")
+                        && (!(this.dialect instanceof PostgresDialect))
                         && (!(this.dialect instanceof MsSqlDialect) || !tableSchemaName.equalsIgnoreCase("dbo"))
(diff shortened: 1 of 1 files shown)


PankajKhairnar07:
@bartoszwalacik Can you please check this PR once

bartoszwalacik:
Hi @PankajKhairnar07 , I will take a look

javers-persistence-sql/src/main/java/org/javers/repository/sql/schema/JaversSchemaManager.java
```diff
@@ -254,6 +254,7 @@ boolean relationExists(String name, String schemaName) {
                 } while(tableSchemaName != null
                         && !tableSchemaName.equalsIgnoreCase("public")
                         && !tableSchemaName.equals("")
+                        && (!(this.dialect instanceof PostgresDialect))
```
bartoszwalacik, line 257: I don't understand this change, which problem is it meant to solve?

bartoszwalacik: @PankajKhairnar07 please explain

bartoszwalacik (changes requested):
please contribute a failing test case to document what's going on, the test should fail when your fix is commented and should pass when it's active
see `PostgreSqlIntegrationWithSchemaTest`

Issue #1249, opened 2022-12-29
Is it possible to add the @DiffIgnore annotation to the beginning of the class in the form of @DiffIgnore("field1", "field2", "field3") instead of specifying it on each individual field?


yunuselci:
While using Javers, I had to use the @DiffIgnore annotation multiple times within a very large form. This led to a lot of repetition for me.

```JAVA
@DiffIgnore("field1","field2")
public class MyClass{
private String field1;
private String field2;
private String field3;
private String field4;
}
```

Is it possible for you to create a new annotation that can be used in this way?"

bartoszwalacik:
no, but you can contribute this feature

bartoszwalacik:
PR merged, waiting to be released in `javers-6.11.0`

PR #1256, opened 2023-01-19
#1249 Adding @DiffIgnoreFields annotation 

EdgarsGarsneks:
Lately I have started to use JaVers and felt that I need to say thank you to this project by making a contribution. 

As mentioned in the #1249 when working with larger entities each property need to be marked with `@DiffIgnore`. There are cases were it is more beneficial to have a list of ignore values provided in annotation on class level. 

As it has been proposed in comments instead of modifying existing annotation, new one has been added `@DiffIgnoreFields`.

It accepts list of property names that will be marked as transient by the `PropertyScanner`. 

Example use case.

```java
@AllArgsConstructor
@DiffIgnoreFields({"a","b"})
public class Myclass {
    private String a;
    private String b;
    private String c;
}
```
Then in main

```java
public static void main(String[] args) {
	Javers javers = JaversBuilder.javers().build();

	Myclass left = new Myclass("1","2","3");
	Myclass right = new Myclass("2","4","6");

	System.out.println(javers.compare(left ,right).getChanges().devPrint());
}
```
Output
```
Changes (1):
commit 0.00 
* changes on com.example.demo.Myclass/ :
  - ValueChange{ property: 'c', left:'3',  right:'6' } 
```
Let me know if more sophisticated testing is required and if I have missed something in the process as I'm new to this code base and tried to stick to current code practices.


bartoszwalacik:
thanks @EdgarsGarsneks , I will do the CR

bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1267
thanks @EdgarsGarsneks for this valueable contribution

PR #1263, opened 2023-02-13
Re-enable db2 support

velo:
I saw it used to be there, then got remove but I have no idea why.

Can we get DB2 back in?


bartoszwalacik:
Hi @velo , Javers never supported DB2, it was just a mapping present in the pollyJDBC lib. Let me know if you are interested in contributing DB2 support in Javers.

bartoszwalacik:
For each DB type we have a dedicated integration test, see `PostgresDockerIntegrationTest`

PR #1261, opened 2023-02-02
Feature: Enable schema management check via config for JaversMongoRepositories

egasimov:
This PR intended to resolve  [Feature: Enable schema management check via config for JaversMongoRepositories #1260](https://github.com/javers/javers/issues/1260)

PR #1264, opened 2023-02-14
Egasimov feature/issue 1260

bartoszwalacik:
from https://github.com/javers/javers/pull/1261


egasimov:
@bartoszwalacik ,
Could you please consider these changes as 6.9.2, rather than 7.. ?

Because my current project which requires above feature was using java 11 and < Spring 3.0.0.

egasimov:
@bartoszwalacik ,
Could you please consider these changes as 6.9.2, rather than 7.. ?

Because my current project which requires above feature was using java 11 and < Spring 3.0.0.

Issue #1252, opened 2023-01-13
Map numeric data value in Object type to proper data type instead of GSON's default double data type


madhavjha-git:
**Is your feature request related to a problem? Please describe.**
GSON by default maps any numeric value defined in Object type to Double. 
Same problem is inherited by Javers as it uses GSON under the hood. 

GSON provided the solution for it long back. Javers should adopt it for better results with shadow api. 
https://github.com/google/gson/pull/1290
 
Example:
```
class TypeValue{
    Integer id;
    Object value;
}
```

> Json: 
>         var jsonStr="{\n" +
>                 "    \"id\":1,\n" +
>                 "        \"value\":9999999999\n" +
>                 "        }";

_**Result without setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)**_
```
        Gson gson= new GsonBuilder()
               // .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
                .create();
        var json=gson.fromJson(jsonStr,TypeValue.class);
```

_{
    "id": 1,
    "value": 9.999999999E9
}_		

_**Result with setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)**_
```
        Gson gson= new GsonBuilder()
                .setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE)
                .create();
      var json=gson.fromJson(jsonStr,TypeValue.class);
```
_{
    "id": 1,
    "value": 9999999999
}_	

**Describe the solution you'd like**
Adding this should solve it. 

    public JsonConverterBuilder() {
        this.gsonBuilder = new GsonBuilder();
        this.gsonBuilder.setExclusionStrategies(new SkipFieldExclusionStrategy());
  

>       this.gsonBuilder.setObjectToNumberStrategy(ToNumberPolicy.LONG_OR_DOUBLE);

        registerBuiltInAdapters(Java8TypeAdapters.adapters());
        registerBuiltInAdapters((List)UtilTypeCoreAdapters.adapters());
        registerJsonAdvancedTypeAdapter(new OptionalTypeAdapter());
    }

**Describe alternatives you've considered**

**Additional context**

bartoszwalacik:
Hi @madhavjha-git, feel free to contribute a PR

madhavjha-git:
Hi @bartoszwalacik  raised PR https://github.com/javers/javers/pull/1280 for the same.

[...]

PR #1280, opened 2023-03-19
Issue#1252 Object to number strategy fix

madhavjha-git:
By default GSON maps object to number in double format. Changed this default behavior by changing strategy to ToNumberPolicy.LONG_OR_DOUBLE. This will enable GSON to map objects to number properly.  See https://github.com/javers/javers/issues/1252 for example.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
+++ b/javers-core/src/main/java/org/javers/core/json/JsonConverterBuilder.java
@@ -25,6 +25,7 @@ public class JsonConverterBuilder {
     public JsonConverterBuilder() {
         this.gsonBuilder = new GsonBuilder();
         this.gsonBuilder.setExclusionStrategies(new SkipFieldExclusionStrategy());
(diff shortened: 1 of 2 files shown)


bartoszwalacik:
@madhavjha-git is this solution backward compatible? What happens if a new version on Javers deserializes data persisted by the previous versions?

madhavjha-git:
@bartoszwalacik Yes . This issue only happens during deserialization. I saved data using previous version and I see data was saved properly in jv_snapshots. 

  "state": {
    "data": {
      "$numberLong": "99999999999"
    },
    "type": "number"
  },

Only after calling findShadows we get 

        {
            "type": "number",
            "data": 9.9999999999E10
        }

But after this objectToNumberStrategy change, it works fine.

bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1282/commits

bartoszwalacik:
thanks @madhavjha-git , I will release it soon

Issue #1230, opened 2022-10-26
Scanned classes ignore custom value comparators for their properties


catautobox:
**Clear description of my expectations versus reality**
Registering a scanned class holding a certain value type property (for example: BigDecimal) and registering a custom value comparator for that value type (for example BigDecimalComparatorWithFixedEquals) via JaversBuilder should make the scanned class use the custom value comparator for that property.

In reality the scanned class seems to use for its property the JaversTypes registered early from WellKnownValueTypes (for example: BigDecimal), ignoring the custom value comparator registered via JaversBuilder.

The value type itself (here: BigDecimal) seems to hold the correct custom value comparator.
But the value type for the type mapping for the scanned class property does not hold the correct custom value comparator (in fact, in the BigDecimal case it does not hold any custom comparator at all).

**Steps To Reproduce**
I have a **runnable test case ** which isolates the bug and allows Javers Core Team to easily reproduce it. I have pushed this test case to my fork of this repository:

https://github.com/catautobox/javers/blob/master/javers-core/src/test/groovy/org/javers/core/cases/JaversBuilderIgnoresCustomValueComparatorForValueTypeForScannedClass.groovy

**Javers' Version**
6.7.1
Earlier versions are probably affected too.

**Additional context**

If I'm not mistaken, in org.javers.core.JaversBuilder#assembleJaversInstance

`
        addModule(new TypeMapperModule(getContainer()));
`

registers WellKnownValueTypes early,
then

```
        for (Class c : classesToScan){
            typeMapper().getJaversType(c);
        }
```

registers scanned classes (at which point their properties will share JaversTypes and custom comparators already registered at that point ),
then

`
        mapRegisteredClasses();
`

registers actual custom comparators.

Registering a CustomValueComparator for a JaversType replaces the entry in the mappedTypes (here for BigDecimal itself), but the mapping for the scanned class property seems to still use a different, earlier instance of the JaveryType or CustomValueComparator.

The fact, that JaveryType mappings are being replaced makes custom value comparators "disappear" when registering custom value type adapters too:

https://github.com/catautobox/javers/blob/master/javers-core/src/test/groovy/org/javers/core/cases/JaversBuilderIgnoresCustomValueComparatorWhenRegisteringValueTypeAdapter.groovy

[...]

PR #1281, opened 2023-03-21
Bugfix 1230 javers builder and type mapper race conditions

catautobox:
#1230 Bugfix for JaversBuilder and TypeMapper race conditions.

Attempts to overwrite a ClientsClassDefinition will result now in an exception.
When registering a value type adapter or a value GSON type adapter, reuse exiting ClientsClassDefinition(s) if present.
Let TypeMapperEngine (and transitively TypeMapper too) gracefully handle duplicate type twice by ignoring the second attempt.
Add exception as sanity check to TypeMapperEngine.addFullMapping() if internally an attempt to overwrite an existing type would be made.
Let JaversBuilder handle user-registered clientsClassDefinitions first to give the client an opportunity to define own ClientsClassDefinition / types / custom value comparators etc. which, in combination of ignoring duplicates in the TypeMapperEngine, will give his types precedence over types introduced by class scanning or core type registration.

All tests executed via:
`./gradlew test`
pass except for:
```
Condition failed with Exception:

calculateToString(ZoneInfo.getTimeZone("UTC")) == "sun.util.calendar.ZoneInfo[id=\"UTC\",offset=0,dstSavings=0,useDaylight=false,transitions=0,lastRule=null]"
                  |
                  java.lang.IllegalAccessError: class org.javers.core.metamodel.type.ValueTypeTest (in unnamed module @0x336eb6ed) cannot access class sun.util.calendar.ZoneInfo (in module java.base) because module java.base does not export sun.util.calendar to unnamed module @0x336eb6ed
                  	at org.javers.core.metamodel.type.ValueTypeTest.should calculate reflective toString(ValueTypeTest.groovy:42)

	at org.javers.core.metamodel.type.ValueTypeTest.should calculate reflective toString(ValueTypeTest.groovy:42)
Caused by: java.lang.IllegalAccessError: class org.javers.core.metamodel.type.ValueTypeTest (in unnamed module @0x336eb6ed) cannot access class sun.util.calendar.ZoneInfo (in module java.base) because module java.base does not export sun.util.calendar to unnamed module @0x336eb6ed
	... 1 more
```
which seems to fail on my machine without my changes too.


javers-core/src/test/groovy/org/javers/core/metamodel/type/TypeMapperTest.groovy
```diff
@@ -279,6 +279,7 @@ class TypeMapperTest extends Specification {
     @Unroll
     def "should ignore all props of subclass when superclass #classType has Included properties definition"(){
         given:
+        def mapper = JaversTestBuilder.javersTestAssembly().typeMapper
```
catautobox, line 282: The new instance for each internal test case was initially necessary because I temporarily made TypeMapperEngine throw on second registration attempt.
The TypeMapperEngine in this branch will ignore the second attempt of registering a definition for PropsClass and the test would pass, but more or less by accident, so I thought I leave the local instance here to make it clear, that the internal test cases expect fresh instances.

Just let me know if I should restore this test to its original state.

bartoszwalacik:
thanks @catautobox , I will continue here https://github.com/javers/javers/pull/1283

bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1283

PR #1290, opened 2023-05-23
When entity has attribute of type coordinate then an attempt to create a commit throws NaN is not a valid double value as per JSON specification

maciej-arkit:


bartoszwalacik:
@maciej-arkit please elaborate what's going on here, try to follow https://github.com/javers/javers/blob/master/CONTRIBUTING.md
Definetly, we don't want to add dependencies to random libraries like org.locationtech.jts

Issue #1306, opened 2023-07-11
Set properties as shallow references via EntityDefinitionBuilder


cberes:
Hi there, I have a feature request. It would be helpful for us if we could set properties as shallow references via EntityDefinitionBuilder. Similar to placing `@ShallowReference` on a field instead of a class. For example, this might look like this

    EntityDefinitionBuilder.entityDefinition(Student.class)
        .withIdPropertyName("id")
        .withShallowProperties(List.of("professor", "school"))
        .build()

Rather than add `@ShallowReference` to many fields in many entities, configuring this via EntityDefinitionBuilder would take only a few lines.

As as an alternative, some way to substitute our own annotation for `@ShallowReference` would work. For example, we want all relationships between entity classes to be shallow. So if there was some way to tell Javers that all fields with our own `@Relationship` annotation are shallow, that would work.

I have a pull request I will add shortly, but I'm sure it will need work.

bartoszwalacik:
Hi @cberes , seems like a good idea, I will take a look at your PR

bartoszwalacik:
released in 7.2.0

PR #1307, opened 2023-07-11
#1306 set shallow reference properties via EntityDefinitionBuilder

cberes:
Hi there, this is a pull request for #1306. I'd be happy to address any feedback or start over if this solution is not acceptable.


bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1309

PR #1354, opened 2024-01-18
add terminalsnapshot configuration and keep terminalsnapshot when it is enabled

yezhongqiang:
refer to https://stackoverflow.com/questions/38696077/how-to-retrieve-deleted-record.

then with code snippet:
javersBuilder.withTerminalSnapshot(true);

queryBuilder.withSnapshotType(SnapshotType.TERMINAL);
javers.findShadows(queryBuilder.build());

to easily get the deleted records.

### DIFF
--- a/javers-core/src/main/java/org/javers/core/CoreConfiguration.java
+++ b/javers-core/src/main/java/org/javers/core/CoreConfiguration.java
@@ -22,6 +22,7 @@ public class CoreConfiguration {
     private final boolean initialChanges;
 
     private final boolean terminalChanges;
+    private final boolean terminalSnapshot;
 
     private final boolean usePrimitiveDefaults;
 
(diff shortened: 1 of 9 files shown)


bartoszwalacik:
Hi @yezhongqiang ,
I understand your requirement but I don't like the idea of copying the full state of a previous snapshot to the terminal snapshot. Not only because of data duplication but also because of lack of backward compatibility.

Please consider using the same approach like in `findChanges`, where a final change set (delete) is calculated as diff between the two last snapshots:

```
    @ToString
    class Entity {
        @Id
        UUID id
        String val
    }

    def "should show full state of deleted object" () {
        given:
        def javers = JaversBuilder.javers().build()
        def entity = new Entity(id:UUID.randomUUID(), val:"abc")

        when:
        javers.commit("author", entity)
        javers.commitShallowDelete("author", entity)
        def query = javers.findChanges(QueryBuilder.byInstance(entity).build())

        println(query.prettyPrint())

        then:
        true
    }
```

```
20:16:28.859 [main] INFO  org.javers.core.Javers - Commit(id:1.00, snapshots:1, author:author, changes - InitialValueChange:2 NewObject:1), done in 62 ms (diff:49 ms, persist:13 ms)
20:16:28.870 [main] INFO  org.javers.core.Javers - Commit(id:2.00, snapshots:1, author:author, changes - ObjectRemoved:1)
Changes:
Commit 2.00 done by author at 11 Feb 2024, 20:16:28 :
* object removed: org.javers.core.cases.Case441UUID$Entity/4b3fbc4f-d585-4fb1-ae29-179f1afbd6af
  - 'id' value '4b3fbc4f-d585-4fb1-ae29-179f1afbd6af' unset
  - 'val' value 'abc' unset
Commit 1.00 done by author at 11 Feb 2024, 20:16:28 :
* new object: org.javers.core.cases.Case441UUID$Entity/4b3fbc4f-d585-4fb1-ae29-179f1afbd6af
  - 'id' = '4b3fbc4f-d585-4fb1-ae29-179f1afbd6af'
  - 'val' = 'abc'
```

bartoszwalacik (changes requested):
see the last comment

Issue #1090, opened 2021-04-23
NavigableMap and TreeMap are not mapped to Map type


dorpsidioot:
**Clear description of my expectations versus reality**
I've got a class which uses NavigableMap on a field level. If I want to detect changes on it, javers throws an error stating:

JaversException MANAGED_CLASS_MAPPING_ERROR: given javaClass 'class java.util.TreeMap' is mapped to MapType, expected ManagedType

But there is no way to indicate via the javers builder that SortedMap/NavigableMap can be seen as instances of Map<>.
If I change my code to Map<K,V> then it works, but I can't change it, because we need the type to be at least of type NavigableMap.

How could I fix this?

**Steps To Reproduce**
Create a class with a property of type NavigableMap<K, V> and add some data to an instance of it. Compare the two objects and the compare method will throw an error.

**Javers' Version**
6.0.1

**Additional context**
It works when I ignore the fields...but I don't want to ignore them. I have a method to return the values as a List, but I can't indicate to javers that it needs to take that method instead of the property.

bartoszwalacik:
@dorpsidioot I will try to fix it, but please write a test case. See https://github.com/javers/javers/blob/master/CONTRIBUTING.md#guidelines-for-bug-reporting

bartoszwalacik:
@hwhh , see my previous comment, for now this bug report is not confirmed , a test case is missing

spaisey:
FWIW, I had exactly the same error message with a class that defined a `SortedMap<>` using an older version of Javers (6.6.5 - although I tried with 7.3.8 and experienced the same).
The problem manifested itself when I upgraded from JDK 17 -> 21.
However, simply registering the `SortedMap` when creating the Javers instance fixed the issue. 

```
JaversBuilder
            .javers()
            .registerValue(SortedMap.class)
            .registerValue(Double.class, new CustomDoubleComparator(serializationProperties.getDecimalPlaces()))
            .build();
```

bartoszwalacik:
might be fixed by https://github.com/javers/javers/pull/1369

bartoszwalacik:
@dorpsidioot please retest it on Javers 7.4.2

Issue #916, opened 2019-11-20
ReflectionUtil.calculateHierarchyDistance does not return all superinterfaces


jeremy-im:
A field of type `kotlinx.collections.immutable.PersistentList` will end up causing "JaversException MANAGED_CLASS_MAPPING_ERROR" because `TypeMapper.findPrototype` is not provided `java.util.List` in the result of `calculateHierarchyDistance`:

- interface `PersistentList` extends `ImmutableList`
- interface `ImmutableList` extends `java.util.List`
- but `calculateHierarchyDistance` does not recursively follow `current.getInterfaces()`, so it only ever returns direct superinterfaces.

Attached a minimal Gradle project (without the gradlew), and the output from executing `gradle run`:

- [PersistentListBug.zip](https://github.com/javers/javers/files/3871377/PersistentListBug.zip)
- [gradle.run.txt](https://github.com/javers/javers/files/3871398/gradle.run.txt)

bartoszwalacik:
Hi Jeremy, few things.

* Javers doesn't have official support for Kotlin, but you can contribute it
* we are not using zip here, follow the guidliness in https://github.com/javers/javers/blob/master/CONTRIBUTING.md and push a failing test to your fork

jeremy-im:
Hi!

This bug happened to be discovered in Kotlin, but it affects all JVM languages. The [classfile spec in section 4.1](https://docs.oracle.com/javase/specs/jvms/se8/html/jvms-4.html#jvms-4.1) defines the `interfaces[]` to contain the direct superinterfaces only. And that is what the reflection API returns via `Class.getInterfaces()`.

Everyone is allowed to implement the java.util collections interfaces using whatever interface hierarchy and programming language they want to. But it must conform to the classfile spec. Javers could easily support all those implementations out of the box if only it recursively followed the superinterfaces (the way it already follows superclasses).

I have provided the best bug report I can given the amount of time I have available to spend on this. If that is not enough for you to work with, then that is unfortunate.

bartoszwalacik:
@jeremy-im this is the open source project. You are supposed to provie a high quality issues and also you are encouraged to contribute bug fixes. 

Bug Reports have to contain:

*    Clear description of your expectations versus reality
*    [Runnable test case](https://github.com/javers/javers/tree/master/javers-core/src/test/groovy/org/javers/core/cases) which isolates the bug and allows us to easily reproduce it on our laptops. You can push this test case to your fork of this repository.

[...]

PR #1369, opened 2024-03-22
ReflectionUtil.calculateHierarchyDistance should return superinterfaces

SimoneGiusso:
This is related to this [issue](https://github.com/javers/javers/issues/916). Actually we noticed this problem when we upgraded our project from Java 17 to Java 21. 

In fact the call to `Javers#compare` on objects which have a property of Type `SortedMap` (assigned to `TreeMap` concrete class) fails in Java 21 with an error similar to:

> JaversException MANAGED_CLASS_MAPPING_ERROR: given javaClass ‘class java.util.TreeMap’ is mapped to MapType, expected ManagedType

While in Java 17, it works. 

The reason is that the mapping between `SortedMap` and JaversType changes when using the two Java versions:

- in Java 17 `SortedMap` is mapped to `MapType` (not a `ManagedType`) ✅
- in Java 21 `SortedMap` is mapped to `ValueObjectType` (subclass of `ManagedType`) ❌

**This is the issue `SortedMap` is not correctly mapped when using Java 21.** 

Full stack error:
> Caused by: org.javers.common.exception.JaversException: MANAGED_CLASS_MAPPING_ERROR: given javaClass ‘class java.util.TreeMap’ is mapped to MapType, expected ManagedType
> 	at org.javers.core.metamodel.type.TypeMapper.getJaversManagedType(TypeMapper.java:178)
> 	at org.javers.core.metamodel.type.TypeMapper.getJaversManagedType(TypeMapper.java:161)
> 	at org.javers.core.metamodel.object.GlobalIdFactory.createId(GlobalIdFactory.java:45)
> 	at org.javers.core.graph.LiveCdoFactory.create(LiveCdoFactory.java:37)
> 	at org.javers.core.graph.EdgeBuilder.buildSingleEdge(EdgeBuilder.java:33)
> 	at org.javers.core.graph.ObjectGraphBuilder.buildSingleEdges(ObjectGraphBuilder.java:107)
> 	at org.javers.core.graph.ObjectGraphBuilder.buildEdges(ObjectGraphBuilder.java:97)
> 	at org.javers.core.graph.ObjectGraphBuilder.buildGraphFromCdo(ObjectGraphBuilder.java:65)
> 	at org.javers.core.graph.ObjectGraphBuilder.buildGraph(ObjectGraphBuilder.java:54)
> 	at org.javers.core.graph.LiveGraphFactory.createLiveGraph(LiveGraphFactory.java:38)
> 	at org.javers.core.diff.DiffFactory.buildGraph(DiffFactory.java:102)
> 	at org.javers.core.diff.DiffFactory.compare(DiffFactory.java:55)
> 	at org.javers.core.JaversCore.compare(JaversCore.java:176)
 
P.S. The message exception, as it is written,  is due to the fact that, at first the `SortedMap` property is considered as `ManagedType` as explained above. Therefore the call to `ObjectGraphBuilder.buildSingleEdges(ObjectGraphBuilder.java:107)` with this property. However since the instance is of `TreeMap`, when retrieving the `JaversManagedType` at `TypeMapper.getJaversManagedType(TypeMapper.java:170`) ,right before the exception, it returns `MapType`. 

It may be related to the following issue as well:

https://github.com/javers/javers/issues/1090

### DIFF
--- a/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
+++ b/javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
@@ -239,10 +239,17 @@ public class ReflectionUtil {
                 parents.add(current);
             }
 
-            for (Class i : current.getInterfaces()) {
-                if (!interfaces.contains(i)) {
-                    interfaces.add(i);
+            Class<?>[] currentInterfaces = current.getInterfaces();
+            while (currentInterfaces.length != 0) {
+                for (final Class<?> i : currentInterfaces) {
+                    if (!interfaces.contains(i)) {
+                        interfaces.add(i);
+                    }
                 }
+
+                currentInterfaces = Arrays.stream(currentInterfaces)
+                        .flatMap(currentInterface -> Arrays.stream(currentInterface.getInterfaces()))
+                        .toArray(Class<?>[]::new);
             }
 
             current = current.getSuperclass();
--- a/javers-core/src/test/groovy/org/javers/common/reflection/ReflectionUtilTest.groovy
+++ b/javers-core/src/test/groovy/org/javers/common/reflection/ReflectionUtilTest.groovy
@@ -66,7 +66,7 @@ class ReflectionUtilTest extends Specification {
 
     def "should calculate hierarchy distance as follows (parents first, interfaces last)"() {
         expect:
-        ReflectionUtil.calculateHierarchyDistance(HashMap) == [AbstractMap, Map, Cloneable, Serializable]
+        ReflectionUtil.calculateHierarchyDistance(TreeMap) == [AbstractMap, NavigableMap, Cloneable, Serializable, SortedMap, Map]
     }
 
     def "should get all methods from a given class without inheritance duplicates"(){
--- a/javers-core/src/test/groovy/org/javers/core/cases/Java21SortedMapShouldBeStillMappedToJaversMapType.groovy
+++ b/javers-core/src/test/groovy/org/javers/core/cases/Java21SortedMapShouldBeStillMappedToJaversMapType.groovy
@@ -1,0 +1,26 @@
+package org.javers.core.cases
+
+
+import org.javers.core.JaversBuilder
+import org.javers.core.metamodel.type.MapType
+import spock.lang.Specification
+
+/**
+ * Related also to issue https://github.com/javers/javers/issues/916 which suggest the in fact:
+ * {@link org.javers.common.reflection.ReflectionUtil#calculateHierarchyDistance} should return not only direct
+ * interfaces but superinterfaces as well.
+ * @author simone giusso
+ */
+class Java21SortedMapShouldBeStillMappedToJaversMapType extends Specification{
+
+    def "should map java 21 SortedMap to Javers MapType instead of ValueObjectType"() {
+        given:
+        final def javers = JaversBuilder.javers().build();
+
+        when:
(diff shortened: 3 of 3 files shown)


javers-core/src/main/java/org/javers/common/reflection/ReflectionUtil.java
```diff
@@ -239,10 +239,17 @@ public static List<Type> calculateHierarchyDistance(Class<?> clazz) {
...
+            while (currentInterfaces.length != 0) {
+                for (final Class<?> i : currentInterfaces) {
+                    if (!interfaces.contains(i)) {
+                        interfaces.add(i);
+                    }
                 }
+
+                currentInterfaces = Arrays.stream(currentInterfaces)
+                        .flatMap(currentInterface -> Arrays.stream(currentInterface.getInterfaces()))
+                        .toArray(Class<?>[]::new);
```
SimoneGiusso, line 252: In Java 21 `SortedMap` doesn't directly extend the `Map` interface. Instead it extends `SequencedMap` and `SequencedMap` extends `Map`. Given an interface or class, getInterfaces(), returns only the interfaces directly implemented. A the end of this method the `interfaces` variable should contains `also` the possible super interfaces.

javers-core/src/test/groovy/org/javers/common/reflection/ReflectionUtilTest.groovy
```diff
@@ -66,7 +66,7 @@ class ReflectionUtilTest extends Specification {
 
     def "should calculate hierarchy distance as follows (parents first, interfaces last)"() {
         expect:
-        ReflectionUtil.calculateHierarchyDistance(HashMap) == [AbstractMap, Map, Cloneable, Serializable]
+        ReflectionUtil.calculateHierarchyDistance(TreeMap) == [AbstractMap, NavigableMap, Cloneable, Serializable, SortedMap, Map]
```
SimoneGiusso, line 69: With the old implementation `SortedMap` is not part of the output here.

bartoszwalacik:
@SimoneGiusso thanks for the detailed bug report, I will do CR of your PR

bartoszwalacik:
Unfortunately, I can't confirm your bug report and solution , which version of JVM do you use exactly?

I tested on Corretto 21, and the mapping works as expected, your test is passing on ReflectionUtil from master
Screenshot 2024-03-30 at 18 37 25

SimoneGiusso:
> Unfortunately, I can't confirm your bug report and solution , which version of JVM do you use exactly?
> 
> I tested on Corretto 21, and the mapping works as expected, your test is passing on ReflectionUtil from master
> 
> Screenshot 2024-03-30 at 18 37 25

I'm using `temurin` but it should not make any difference. Can you please make sure you are running Java 21? Try to print out this:

`System.getProperty("java.version")`

bartoszwalacik:
You are right, btw can you give me a hint how to run a test on higher Java version than specified in `build.gradle`?
Whatever I do, Gradle runs a test on JVM version from build.gradle :
```
java {
    toolchain {
        languageVersion = JavaLanguageVersion.of(11)
    }
}
```

bartoszwalacik:
ok, got it

bartoszwalacik:
ok, got it
Screenshot 2024-03-30 at 20 26 13

bartoszwalacik:
merged https://github.com/javers/javers/pull/1371

bartoszwalacik:
the issue is fixed, released in 7.4.2
Thanks @SimoneGiusso !

Issue #1382, opened 2024-06-03
@DiffIgnore Annotation on Subclass Fields Sometimes Ignored


clarkhan:
**Clear description of my expectations versus reality**
The `@DiffIgnore` annotation on fields within subclasses seems to be ignored in certain scenarios. For instance, when retrieving the `TypeMapping` of a superclass before accessing the `TypeMapping` of a subclass, the ignored fields in the subclass are not treated properly.

**Steps To Reproduce**
I have a **runnable test case ** which isolates the bug and allows Javers Core Team to easily reproduce it. I have pushed this test case to my fork of this repository:

https://github.com/clarkhan/javers/blob/master/javers-core/src/test/groovy/org/javers/core/cases/CaseInheritanceWithIgnoredField.groovy

```groovy
package org.javers.core.cases
import jakarta.persistence.Id
import org.javers.core.JaversBuilder
import org.javers.core.metamodel.annotation.DiffIgnore
import org.javers.core.metamodel.annotation.TypeName
import org.javers.core.metamodel.type.EntityType
import spock.lang.Specification

class CaseInheritanceWithIgnoredField extends Specification {
    @TypeName("A")
    class A {
        @Id
        String id
        String field
    }

    @TypeName("B")
    class B extends A {
        @DiffIgnore
        String someOtherField
    }

    @TypeName("C")
    class C extends A {
        @DiffIgnore
        String someOtherField
    }

    def "should ignore fields on sub Class that are annotated with DiffIgnore"(){
        given:
        def javers = JaversBuilder.javers().build()

        when:
        EntityType bType = javers.getTypeMapping(B)
        EntityType aType = javers.getTypeMapping(A)
        EntityType cType = javers.getTypeMapping(C)

        then:
        assert bType.findProperty("someOtherField").isEmpty() // passed
        assert cType.findProperty("someOtherField").isEmpty() // failed
    }
}
```

result:
```text
EntityType{
  baseType: class org.javers.core.cases.CaseInheritanceWithIgnoredField$A
  typeName: A
  managedProperties:
    Field id String, javersType: ValueType, declared in: A
    Field field String, javersType: ValueType, declared in: A
  idProperties: [id]
}
EntityType{
  baseType: class org.javers.core.cases.CaseInheritanceWithIgnoredField$B
  typeName: B
  managedProperties:
    Field id String, javersType: ValueType, declared in: A
    Field field String, javersType: ValueType, declared in: A
  idProperties: [id]
}
EntityType{
[...]

PR #1387, opened 2024-06-05
fix for https://github.com/javers/javers/issues/1382

clarkhan:


bartoszwalacik:
fix for https://github.com/javers/javers/issues/1382

javers-core/src/test/groovy/org/javers/core/cases/CaseInheritanceWithIgnoredField.groovy
```diff
@@ -0,0 +1,50 @@
...
+        @Id
+        String id
+        String field
+    }
+    @TypeName("B")
+    class B extends A {
+        @DiffIgnore
+        String someOtherField
+    }
+    @TypeName("C")
```
bartoszwalacik, line 25: not sure what's the difference between B and C. Do we need both these classes in this test?

bartoszwalacik: ok, the order of calls to getTypeMapping() is importany, got it

bartoszwalacik:
merged from here https://github.com/javers/javers/pull/1388

Issue #1400, opened 2024-07-30
ClassCastException: MissingProperty cannot be cast to Optional


bartebor:
When comparing two objects with different `Optional` fields, an exception is thrown:
```
java.lang.ClassCastException: class org.javers.core.metamodel.property.MissingProperty cannot be cast to class java.util.Optional (org.javers.core.metamodel.property.MissingProperty is in unnamed module of loader 'app'; java.util.Optional is in module java.base of loader 'bootstrap')
      at org.javers.core.metamodel.type.OptionalType.map(OptionalType.java:43)
      at org.javers.core.graph.LiveNode.getDehydratedPropertyValue(LiveNode.java:112)
      at org.javers.core.diff.NodePair.getRightDehydratedPropertyValueAndSanitize(NodePair.java:135)
      at org.javers.core.diff.appenders.OptionalChangeAppender.calculateChanges(OptionalChangeAppender.java:33)
      at org.javers.core.diff.DiffFactory.appendChanges(DiffFactory.java:165)
      at org.javers.core.diff.DiffFactory.appendPropertyChanges(DiffFactory.java:155)
      at org.javers.core.diff.DiffFactory.createAndAppendChanges(DiffFactory.java:137)
      at org.javers.core.diff.DiffFactory.create(DiffFactory.java:70)
      at org.javers.core.diff.DiffFactory.compare(DiffFactory.java:55)
      at org.javers.core.JaversCore.compare(JaversCore.java:176)
```
**Steps To Reproduce**
Following code triggers this problem:
```groovy
    static class Container {
        final String id;
        final Base base;

        Container(String id, Base base) {
            this.id = id;
            this.base = base;
        }
    }

    static abstract class Base {

        static class Concrete1 extends Base {
            final Optional<String> string1;

            Concrete1(String string1) {
                this.string1 = Optional.of(string1);
            }
        }

        static class Concrete2 extends Base {
            final Optional<String> string2;

            Concrete2(String string2) {
                this.string2 = Optional.of(string2);
            }
        }
    }

// ...

        def javers = JaversBuilder.javers()
            .registerEntity(new EntityDefinition(Container.class, "id"))
            .build()

        def c1 = new Container("id", new Base.Concrete1("string1"));
        def c2 = new Container("id", new Base.Concrete2("string2"));

        def changes = javers.compare(c1, c2) // throws
```

[...]

PR #1401, opened 2024-07-31
Test case for #1400

bartoszwalacik:
see https://github.com/javers/javers/issues/1400


bartoszwalacik:
moved here https://github.com/javers/javers/pull/1402

Issue #1415, opened 2024-09-28
Issue with GSON builder  throwing "NaN is not a valid double value as per JSON specification."


BrakefieldMN:
**Clear description of my expectations versus reality**
An object with a double value set as Double.NaN should be persisted. In reality, javers' gson builder throws an error because it does not leverage "serializeSpecialFloatingPointValues". 

**Steps To Reproduce**
I have a **runnable test case ** which isolates the bug and allows Javers Core Team to easily reproduce it. I have pushed this test case to my fork of this repository:

https://github.com/BrakefieldMN/javers/blob/master/javers-spring-boot-starter-mongo/src/test/groovy/org/javers/spring/boot/mongo/JaversMongoStarterNullDoubleTest.groovy

**Javers' Version**
7.6.2

**Additional context**
This issue was identified a few years ago. 
https://stackoverflow.com/questions/54573968/issue-in-infinity-is-not-a-valid-double-value-as-per-json-specification-with-flo 

I also intend to open a PR to make the adjustment in the gson builder.

bartoszwalacik:
Hi @BrakefieldMN , thanks for the bug report. We are happy to see your pull request with a fix

BrakefieldMN:
I tried following the documentation for the process. Please let me know if I missed or need to change anything! Thank you for your response.

bartoszwalacik:
thanks , I will take a look at your PR in a few days

bartoszwalacik:
the fix is released in 7.6.3

PR #1418, opened 2024-10-01
Fixing Issue #1415 with Double.NaN

BrakefieldMN:
This PR resolves the issue found here: #1415

The previously failing test case now correctly persists the object containing Double.NaN.


bartoszwalacik:
build - https://github.com/javers/javers/pull/1420

Issue #1295, opened 2023-06-08
AOT Support?


korkutkose:
As i understood that, currently, neither programatically nor at repository level [https://github.com/oracle/graalvm-reachability-metadata](graalvm-reachability-metadata) JaVers supporting spring-native yet. Is there any plan for this? 
It's cumbersome to extract necessary reflections via tracability agent to be honest since we need to keep hitting necessary endpoints programatically to generate reflection configs.

bartoszwalacik:
sorry, but I have zero knowledge about GraalVM, btw this issue is unclear for me. Is it a question? Is it a feature request? If so, please follow the issue template

**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]

**Describe the solution you'd like**
A clear and concise description of what you want to happen.

**Describe alternatives you've considered**
A clear and concise description of any alternative solutions or features you've considered.

**Additional context**
Add any other context or screenshots about the feature request here.

Once an issue is accepted -- we are open for contributions

goafabric:
@bartoszwalacik @korkutkose 
Maybe i can shed some light here, as i would have been also interested in native support.
For anyone who stumbles upon this issue.

In a nutshell, AOT/Native support, gives you the option to compile native machine code images, from pure java.
While this has several benefits, the downside is that you need to register so called "hints" for everything the compiler know needs ahead of time, like reflection, proxies etc.

That works good if there will be only a few of these hints.
However, i tried to get the ball running with javers .. and unfortunately as amazingly simple javers works,
i ended up with 100+ reflection hints, it's still not working because of a hibernate problem .. at least it know survives the general bootstrap phase.

I will attach the spring boot Configuration for the sake of clearness.

But it think with so much reflection inside javers, aot is not a good fit for now.
At least it helped me to get a good inspiration how to build my own auditing, based an AOP with Spring data, thx 4 that

goafabric:
Native hint config
[JaversNativeConfiguration.java.zip](https://github.com/javers/javers/files/12450824/JaversNativeConfiguration.java.zip)

goafabric:
> @goafabric thanks for explanation, btw if you guys are eager to contribute this kind of support to javers -- let me know

we will and thx .. but usually setting and also maintaining 100+ hints does not make a lot of sense

goafabric:
so for everyone stumbling here i can provide the file below it lakes make everything bootstrap and write
[...]

PR #1421, opened 2024-11-05
Spring Native Support added

goafabric:
This PR adds Spring Boot Native Support
We need to add the Native hints, with reflection during build time due to Classes being package protected.

And we need to update the JpaHibernateConnectionProvider, 
because we will get a Nullpointer on EntityManager otherwise


javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/aot/JaversNativeHintsConfiguration.java
```diff
@@ -0,0 +1,170 @@
...
+        @Override
+        public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+
+            hints.resources().registerResourceBundle("org.aspectj.weaver.weaver-messages");
+
+            var classes = Arrays.asList(
+                    "java.util.ArrayList",
+
+                    // Package: org.javers.core
+                    "org.javers.core.Javers",
```
bartoszwalacik, line 24: 1. hardcoding all class names seems fragile, what if we change something?

2. Could you please explain (or point to some docs) why do we need to register all these names to achieve Spring Native support 

3. how this move affects classic jvm apps?

goafabric: Hi bartoszwalacik,

1.
hardcoding class names is unfortunately exactly how native hints work ... there is basically no other alternative than this.
You need to tell the compiler upfront which classes need to be registered for reflection, so that they are prepared during compilation.
If classes are changed a Runtime Exception will be thrown as the Class.Forname fails then.
Normally classes are listed as Classes, and not as literals, which would automatically change upon Renaming the class.

Unfortunately Javers uses a lot of Package protected classes, that are not importable this way.

2.
See 1. .. the compiler needs to be told upfront which classes use reflection and all of these classes do so.
https://www.baeldung.com/spring-native-intro

3.
It should not affect jvm apps at all

bartoszwalacik: @goafabric thanks for your answer, still not sure I fully understand it, most Javers' classes are , well, just normal classes, only a few of them, or maybe only the one -- ReflectionUtil -- uses reflection API, so why registering them all?

bartoszwalacik: concerning 3. , from what I see, this RuntimeHintsRegistrar is executed unconditionally
so it will slow down starting Javers for normal JVM apps. Could you please come up with some condition, end execute it only when needed by a native image app? might is also be an alternative starter

goafabric: The RuntimeRegistrar is only executed during build time, never during run time.
You may validate that by setting a break point.
And then also only if the plugin "org.graalvm.buildtools.native" is inside the build.

Concerning Reflection, it is less about classes that use ReflectionUtil and more about which classes are accessed reflectively.
This will be every ClassForname(class) and every field, method, constructor etc that will be executed via reflection.

The list was initially generated the graalvm agent.
And after that you have to add manually further classes the agent did not catch.

--- example---
class A() {
  doReflectionStuff() {
     class.forName(B.class);
     class.forName(C.class);
     class.forName(D.class);
  }
}

=> needs reflection hints for B,C + D, not for A

bartoszwalacik: got it, thanks for the explanation, last thing before merge, could you please add this AOT support also the the javers Mongo starter? (see https://github.com/javers/javers/blob/master/javers-spring-boot-starter-mongo/src/main/java/org/javers/spring/boot/mongo/JaversMongoAutoConfiguration.java)

bartoszwalacik:
builds are visible here https://github.com/javers/javers/compare/master...spring-boot-native-configuration

javers-spring-boot-starter-sql/src/main/java/org/javers/spring/boot/aot/JaversNativeHintsConfiguration.java
goafabric: thank you .. i had the same idea, but unfortunately it's not so easy done, 
as i would have to do them process of figuring out the right hints for mongodb again .. don't think this will be 1:1 copy ..
so my idea was to first get the JPA stuff merged and then eventually taking a second look when I find some time

bartoszwalacik: thanks, maybe we can move this large array of class names to the javers-spring package and share for two starters (no duplication then). and just a question, how can I refresh this array of class names when sth changes?

goafabric: ok I took the extra time for the refactoring .. I hope we can now getting it merged ...

bartoszwalacik: thanks @goafabric , well done, I will merge and release soon

bartoszwalacik:
related issue https://github.com/javers/javers/issues/1295

bartoszwalacik:
released in 7.7.0

PR #1454, opened 2025-09-14
Fixes #1370: Fixes GlobalID hash collission.

### LINKED ISSUE #1370: Invalid Diff because of duplicate hashes of global ids
Comparing two collections by using `<T> Diff compareCollections(Collection<T> oldVersion, Collection<T> currentVersion, Class<T> itemClass);`.
The elements in the collection are the same type.
The elements in the collection are entities.

Expectation:
All objects, identified by global id, that are present in the `currentVersion`, but missing in `oldVersion`, are present in the `NewObject` changes.

Reality:
Some entites, that are new, to not produce a `NewObject` change.

...

**Steps To Reproduce**
1. Create two collections.
2. Add elements to the collection with a different ID, that produces the same hash value. e.g. "FB" and "Ea"
3. Compare the two lists by using  `<T> Diff compareCollections(Collection<T> oldVersion, Collection<T> currentVersion, Class<T> itemClass);`

**Javers' Version**
7.3.8

**Additional context**
I identified the root cause of this:
In [GraphPair](https://github.com/javers/javers/blob/0e7ba7a7d52597c55e9ef63d4389f7c39e7f51ed/javers-core/src/main/java/org/javers/core/diff/GraphPair.java#L38) the graphs are built and the comparison in [difference](https://github.com/javers/javers/blob/0e7ba7a7d52597c55e9ef63d4389f7c39e7f51ed/javers-core/src/main/java/org/javers/common/collections/Sets.java#L76) is done by using a hasher.
The wrong result happens because 2 **different** global IDs produce the same hash value.
So the oldVersion contains a **different** element with the same hash, thus it is not recoignized as a new object in the `currentVersion`.

jamnaritesh:
Hi,
Please help to review fix for #1370.

Thanks,

### DIFF
--- a/javers-core/src/main/java/org/javers/common/collections/Sets.java
+++ b/javers-core/src/main/java/org/javers/common/collections/Sets.java
@@ -73,7 +73,7 @@ public class Sets {
     /**
      * null args are allowed
      */
-    public static <E> Collection<E> difference(Set<E> first, Set<E> second, Function<E, Integer> hasher) {
+    public static <E> Collection<E> difference(Set<E> first, Set<E> second, Function<E, String> hasher) {
         if (first == null || first.size() == 0) {
             return Collections.emptyList();
         }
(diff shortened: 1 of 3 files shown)


bartoszwalacik:
Hi @jamnaritesh , well done, thanks!

jamnaritesh:
@bartoszwalacik when can we expect a release next? Sorry i couldn't find the release schedule anywhere.

Issue #1476, opened 2026-03-30
Supporting ignoreCase for withChangedProperty(String property)


D2003ROB:
**Problem:** The framework already supports searching by changed properties, but it currently **requires** exact casing. It would be more user-friendly to allow case-insensitive matching so that partial or lowercase inputs still return relevant results. For example, a field like `doorAppAccess` should be matched when a user searches for `doorapp`, whereas at the moment it only matches if the input is `doorApp`.

**Solution:** A possible approach would be to apply `LOWER()` to the `changed_properties` column and convert the search input to lowercase as well. For example, the current implementation is:
https://github.com/javers/javers/blob/22142f7121a47349a1d6c1c48e585434df77941d/javers-persistence-sql/src/main/java/org/javers/repository/sql/finders/SnapshotQuery.java#L70-L71

This can be adapted to:
`.map(it -> "LOWER(SNAPSHOT_CHANGED) LIKE '%" + it.toLowerCase() + "%'")`

**What I'm asking:** Please advise whether this approach is appropriate. It appears simple to implement, but I assume there may be limitations or trade-offs that are not immediately obvious, given that it has not been implemented already. If it's fine, I can proceed with an implementation.

bartoszwalacik:
Hi, you mean applying LOWER() on both sided in query, or writing with LOWER and then only on one side of a query? Did you consider also MongoDB impl? 
Not sure about the perf impact, does this query use any index now?

D2003ROB:
Hello! 

Now that i read this again, using LOWER on both sides **_might_** be redundant. I'd say that's only required for the existing `changed_properties` column's value. So only for the **STORED VALUE**.

I did not think about the MongoDB implementation, I'll be transparent I haven't used Mongo before, but I'm happy to implement this feature there too. 

On the SQL side, i don't see any indexes that would help this. 

However, on the Mongo side, i see this
https://github.com/javers/javers/blob/22142f7121a47349a1d6c1c48e585434df77941d/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoSchemaManager.java#L58
So I think Mongo might use this index in the query.

bartoszwalacik:
Imo it should be added on both sizes , otherwise it could break existing queries.

For example , existing query : 

cp.changed_properry = 'myProperty'

Breaking:

LOWER(cp.changed_property) = 'myProperty'

PR #1477, opened 2026-04-05
FEATURE: Make withChangedProperty search case-insensitive

D2003ROB:
It is now possible to search within the `changed_properties` array using case-insensitive query. For example, if a property is named `tradingAppAccess`, a query like `tradingap` will return a match. Previously, only inputs matching the exact casing pattern, such as `tradingAp`, would work.

Issue: #1476

### DIFF
--- a/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
+++ b/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
@@ -185,9 +185,14 @@ public class MongoRepository implements JaversRepository, ConfigurationAware {
 
     private Bson createIdQueryWithAggregate(GlobalId id) {
         if (id instanceof InstanceId) {
+            Object cdoId = ((InstanceId)id).getCdoId();
+            if (cdoId instanceof UUID) {
+                UUID uuidCdoId = (UUID) cdoId;
+                cdoId = uuidCdoId.toString();
+            }
             return Filters.or(
                 createIdQuery(id),
-                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, ((InstanceId)id).getCdoId())
+                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, cdoId)
             );
         }
 
--- a/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
+++ b/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
@@ -343,8 +343,13 @@ public class MongoRepository implements JaversRepository, ConfigurationAware {
                 query = addCommitPropertiesLikeFilter(query, params.commitPropertiesLike());
             }
             if (params.changedProperties().size() > 0) {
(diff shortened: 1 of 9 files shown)


javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
```diff
@@ -343,8 +343,13 @@ private Bson applyQueryParams(Bson query, Optional<QueryParams> queryParams) {
                 query = addCommitPropertiesLikeFilter(query, params.commitPropertiesLike());
             }
             if (params.changedProperties().size() > 0) {
-                query = Filters.and(query, Filters.or(params.changedProperties().stream()
-                        .map(it -> new BasicDBObject(CHANGED_PROPERTIES, it)).collect(Collectors.toList())));
+                List<Bson> regexFilters = params.changedProperties().stream()
```
D2003ROB, line 346: `(?)` I'm not sure this is the right approach.. As i see Mongo uses an index, and I'm thinking this way the index will be ignored and mongo will use a collation scan. Any opinions? 

https://github.com/javers/javers/blob/22142f7121a47349a1d6c1c48e585434df77941d/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoSchemaManager.java#L58

`(💡)` Perhaps leaving this unchanged and introducing a new query parameter, `changedPropertiesIgnoreCase`, would be a better approach. This makes sure the existing implementation is not disturbed, while the new case-insensitive option operates independently and does not rely on an index.

bartoszwalacik: Right, the index will be ignored, also it would change the begaviour of existing queries. So definetly a new method , `changedPropertiesIgnoreCase()`, with documented perf impact in the javadoc, is a way to go

bartoszwalacik: more details about this by Claude:
```
1. CRITICAL: Breaking change — no opt-in mechanism

 The PR changes the default behavior of withChangedProperty() from case-sensitive (exact match) to case-insensitive. This is a breaking
 change for all existing users. Java property names are inherently case-sensitive (name != Name), and the current exact-match behavior is
 correct for the standard use case.

 Recommendation: Add a new method withChangedPropertyIgnoreCase(String) (and withChangedPropertyInIgnoreCase(String...)) instead of
 modifying the existing ones. This preserves backward compatibility. Requires a new boolean flag changedPropertiesIgnoreCase flowing through
  QueryParams.

 2. CRITICAL: MongoDB — regex injection vulnerability

 MongoRepository.java builds a regex from user input without escaping:
 new BasicDBObject("$regex", ".*" + it.toLowerCase(Locale.ROOT) + ".*")
 If it contains regex metacharacters (e.g. "foo.*bar"), it will be interpreted as a regex pattern. The input must be escaped with
 Pattern.quote() or equivalent.

 3. CRITICAL: MongoDB — performance regression (always-on)

 The PR replaces an exact-match query (which uses the changedProperties index) with a $regex + $options: "i" query that cannot use indexes,
 forcing collection scans. Since this is now the default (not opt-in), ALL existing withChangedProperty queries will regress in performance.
 ```

PR #1479, opened 2026-05-20
#1373 fix for withChildValueObjects() if @id has java.util.UUID type when using MongoRepository

### LINKED ISSUE #1373: withChildValueObjects() does not work if @id has java.util.UUID type with mongo database
**Clear description of my expectations versus reality**
I noticed that the query with withChildValueObjects() stops returning all child Value Objects owned by selected Entities. This works correctly in javers 7.3.8. I found out that the UUID type is stored as a string in mongo database but [here](https://github.com/javers/javers/commit/8b121a3ed0b474119e12c5c99e9c8fec64509357#diff-1e26833a4a58634b1b88c55d45a9b01dbfd4e3a7f5c7b1ce8af78da756e31bafR188) the java.util.UUID type is used instead of string

**Steps To Reproduce**
The following test works fine in 7.3.8 but fails in 7.4.2
```java
  class Parent {
    @Id
    UUID id;
    List<Child> children;
  }
  
  class Child {
    List<GrandChild> grandChildren;
  }
  
  class GrandChild {
    String value;
  }
  
  def "withChildValueObjects() should work with UUID as entity id"() {
    given:
    GrandChild grandChild1 = new GrandChild(value: "value1")
    GrandChild grandChild2 = new GrandChild(value: "value2")
    GrandChild grandChild3 = new GrandChild(value: "value3")
    GrandChild grandChild4 = new GrandChild(value: "value4")
    GrandChild grandChild5 = new GrandChild(value: "value5")
    GrandChild grandChild6 = new GrandChild(value: "value6")
    
    Child child1 = new Child(grandChildren: [grandChild1, grandChild2])
    Child child2 = new Child(grandChildren: [grandChild3, grandChild4, grandChild5])
    Child child3 = new Child(grandChildren: [grandChild6])
    
    Parent parent = new Parent(id: UUID.randomUUID(), children: [child1, child2, child3])

    javers.commit("author", parent)
    
    when:
    def query = byInstance(parent).withChildValueObjects().build()
    def snapshots = javers.findSnapshots(query)
    
    then:
    assert snapshots.size() == 10 // 1x Parent, 3x Child, 6x GrandChild
  }
```

there is a need to define a UUID representation. It can be done in `org.javers.repository.mongo.DockerizedMongoContainer`:
```java
    DockerizedMongoContainer() {
        this.mongoDBContainer = startMongo()
        this.mongoClient = MongoClients.create(MongoClientSettings
                .builder()
                .uuidRepresentation(UuidRepresentation.STANDARD)
                .applyConnectionString(new ConnectionString(mongoDBContainer.replicaSetUrl))
                .build())
    }
```

**Javers' Version**
7.4.2

**Additional context**
...

jarebudev:
This MR adds a check to `MongoRepository` so that if 'UUID' id types are converted to a `String` when building a query with the cdo id.  This is as the id is stored as a String in `globalId.ownerId.cdoId`, but passing the UUID id object into the Mongo query builder will query for a UUID type.

I've added the E2E test that was documented in the issue in a separate branch to demonstrate the issue before this fix.  This is on this branch https://github.com/javers/javers/compare/master...jarebudev:javers:1373_demonstrate_uuid_error

Fixes #1373

### DIFF
--- a/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
+++ b/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
@@ -185,9 +185,14 @@ public class MongoRepository implements JaversRepository, ConfigurationAware {
 
     private Bson createIdQueryWithAggregate(GlobalId id) {
         if (id instanceof InstanceId) {
+            Object cdoId = ((InstanceId)id).getCdoId();
+            if (cdoId instanceof UUID) {
+                UUID uuidCdoId = (UUID) cdoId;
+                cdoId = uuidCdoId.toString();
+            }
             return Filters.or(
                 createIdQuery(id),
-                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, ((InstanceId)id).getCdoId())
+                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, cdoId)
             );
         }
 
--- a/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
+++ b/javers-persistence-mongo/src/main/java/org/javers/repository/mongo/MongoRepository.java
@@ -185,14 +185,9 @@ public class MongoRepository implements JaversRepository, ConfigurationAware {
 
     private Bson createIdQueryWithAggregate(GlobalId id) {
         if (id instanceof InstanceId) {
-            Object cdoId = ((InstanceId)id).getCdoId();
-            if (cdoId instanceof UUID) {
-                UUID uuidCdoId = (UUID) cdoId;
-                cdoId = uuidCdoId.toString();
-            }
             return Filters.or(
                 createIdQuery(id),
-                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, cdoId)
+                new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, ((InstanceId)id).getCdoId().toString())
             );
         }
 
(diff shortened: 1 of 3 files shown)


bartoszwalacik (changes requested):
Thanks for the fix and the well-structured test!

The root cause analysis is correct — `getCdoId()` returns a `java.util.UUID` object, but MongoDB stores it as a String in `globalId.ownerId.cdoId`, so the BSON type mismatch causes the query to miss results.

However, the `instanceof UUID` check is too narrow. Any non-String, non-primitive `@Id` type could have the same problem. Since the cdoId is always stored as a String in MongoDB, a simpler and more robust fix would be:

```java
new BasicDBObject(GLOBAL_ID_OWNER_ID_CDO_ID, ((InstanceId)id).getCdoId().toString())
```

This way we don't need the `instanceof` check at all — it handles UUID and any other custom Id type in one shot.

jarebudev:
Thanks for reviewing - I've pushed the suggested change to the fix.

bartoszwalacik:
Hey @jarebudev, thanks ,
   I initially suggested using unconditional `.toString()` on `cdoId` (commit a749581), but that approach breaks the existing tests:

   - `should query withChildValueObjects for snapshots and changes by InstanceId`
   - `should query withChildValueObjects for snapshots and changes by Entity type`

   **Why it breaks:** When the `@Id` is an integer type (e.g. `Integer`, `Long`), MongoDB stores `globalId.ownerId.cdoId` as a number
   (`1`). Calling `.toString()` converts it to a string (`"1"`), causing a BSON type mismatch — the aggregate query finds no matches.

   **The fix:** I've reverted to your original approach (commit c0a89a55) — applying `.toString()` only when `cdoId instanceof UUID`. This
    correctly handles the UUID case without breaking integer/long ids. See commit 55e2d9f.

bartoszwalacik:
all good, I will release this

PR #1482, opened 2026-06-03
Fixes #1481: Register core types before classpath scan

### LINKED ISSUE #1481: Built-in ValueTypes are registered after class scanning, causing intermittent InaccessibleObjectException on Java 24+
**Clear description of my expectations versus reality**

### Expectation

Built-in JaVers value types, such as `UUID` and `LocalDateTime` should be registered before package/class scanning begins. When JaVers encounters fields of these types while scanning audited entities, it should immediately recognise them as built-in ValueTypes and avoid any deep introspection of their internal fields.

### Reality

On recent Spring Boot versions running on Java 24+, we occasionally encounter startup failures during JaVers initialisation:

```java
java.lang.reflect.InaccessibleObjectException:
Unable to make field private final long java.util.UUID.mostSigBits accessible:
module java.base does not "opens java.util" to unnamed module @4efc180e
at org.javers.common.reflection.JaversMember.setAccessibleIfNecessary(JaversMember.java:95)
at org.javers.common.reflection.JaversField.<init>(JaversField.java:17)
at org.javers.common.reflection.JaversFieldFactory.createJField(JaversFieldFactory.java:43)
```

This happens while JaVers scans our JPA entities.
After investigating, we found that `JaversBuilder` currently scans classes before registering built-in/core types:

```java
classesToScan.forEach(c -> typeMapper().getJaversType(c));
typeMapper().registerCoreTypes(
    coreConfiguration,
    clientsClassDefinitions.values()
);
```

[Implemented here](https://github.com/javers/javers/blob/aca5060c969ceca474581a60f93d0fb5c6db59b4/javers-core/src/main/java/org/javers/core/JaversBuilder.java#L177) 

This can lead to the following sequence:

- JaVers scans an entity containing a field such as:

```java
private LocalDateTime createdAt;
```

- `LocalDateTime` is not yet registered as a built-in `ValueType`.
- JaVers infers a type mapping for it during scanning.
- The type may be treated as a `ValueObjectType` and introspected.
- Later, `registerCoreTypes(...)` is called, but the type is already registered and is not replaced.
- JaVers attempts reflective access to internal JDK fields.
- On Java 24+, stricter module encapsulation causes an `InaccessibleObjectException`.

We observed the same behaviour with entities using `UUID` identifiers.

**Steps To Reproduce**

Unfortunately, we have not been able to create a deterministic runnable test case.
The issue appears to be highly dependent on:
- the number of scanned entities,
- scan order,
- presence of many `UUID` identifiers,
- presence of many `LocalDateTime` audit fields (`createdAt`, `updatedAt`, etc.),
- Java 24+ module access restrictions.

In our application, the probability increased significantly as the number of audited JPA entities grew.
Although we cannot provide a deterministic reproduction, we were able to eliminate the issue completely by modifying the initialisation order in our JaVers fork.
We changed `JaversBuilder` so that core types are registered before class scanning:

```java
typeMapper().registerCoreTypes(
    coreConfiguration,
    clientsClassDefinitions.values()
);
classesToScan.forEach(c -> typeMapper().getJaversType(c));
```

Since applying this change, we have not observed the exception again.

**Javers' Version**

**7.9.0** 

**Additional context**

Our environment:

- Spring Boot: `3.5.8`
- Java: 24+
- MongoDB
- Large number of audited JPA entities
- Most entities contain:
  - `UUID` identifiers
  - `LocalDateTime createdAt`
  - `LocalDateTime updatedAt`

We suspect the root cause is the initialisation order inside `JaversBuilder`.
Registering built-in/core types before scanning appears to prevent:

- incorrect early type inference,
- unnecessary introspection of JDK value types,
- reflective access warnings/errors on modern JVMs,
- inconsistent snapshot behaviour for audit fields.

If useful, I can also provide the fork/commit containing the change that resolved the issue for us.

pastouret:
Closes #1481 reorder instrumentation of classes scan to ensure built-in types are registered first

### DIFF
--- a/javers-core/src/main/java/org/javers/core/JaversBuilder.java
+++ b/javers-core/src/main/java/org/javers/core/JaversBuilder.java
@@ -169,12 +169,14 @@ public class JaversBuilder extends AbstractContainerBuilder {
 
         bootDateTimeProvider();
 
+        // IMPORTANT: Register core/built-in types BEFORE scanning packages
+        // This ensures LocalDateTime, UUID, etc. are recognized as Value types
+        // before they're encountered in user classes
+        typeMapper().registerCoreTypes(coreConfiguration, clientsClassDefinitions.values());
+
         // classes to scan & additionalTypes
         classesToScan.forEach(c -> typeMapper().getJaversType(c));
         typeMapper().addPluginTypes(additionalTypes);
-
-        // register core / well known types last
-        typeMapper().registerCoreTypes(coreConfiguration, clientsClassDefinitions.values());
 
         // client definitions (explicit types) should have the highest priority
         // if a client would like to register overwrite core types with


bartoszwalacik:
thanks, I will CR on Monday

bartoszwalacik (approved):
Thanks for the excellent bug report and the fix, @pastouret! 👍

The analysis is spot on — the root cause traces back to #1230 which moved `registerCoreTypes()` after class scanning. Before that, core types were registered in the `TypeMapperEngine` constructor, so they were always available during scanning.

The non-deterministic nature of the bug comes from the `votes`/hints mechanism in `TypeFactory` — when an Entity with a UUID `@Id` happens to be scanned first, it saves an `EntityIdHint` that causes UUID to be correctly inferred as `ValueType`. But with `HashSet` iteration order being unpredictable, sometimes UUID gets inferred before any hint is saved, falls through to `createDefaultType()`, and gets treated as a `ValueObject` — triggering deep introspection of JDK internals.

The fix is safe because `registerCoreTypes()` already receives `clientsClassDefinitions` as `classesToSkip`, so user-defined type overrides are preserved regardless of call order.

Merging!

Issue #1484, opened 2026-06-12
Javers detects changes when comparing complex but equal objects


Adrien-dev25:
## Reminder 

**Sets are unordered collections** : By definition, the order of elements in a `Set` should not affect Java equality. 

## Expectation 

Javers is able to detect changes between complex Objects containing other Objects or `Set`. 

1. I expect the method `Javers.compare()` also **respect the above property** : that is to say detect no change between two `Set` having same elements but just in a different order. 

2. I expect the method `Javers.compare()` respect the above property **whatever the "complexity" of compared objects**.

From a general point of view : 
If I have `Objects.equals(object1, object2)` then I expect `JAVERS.compare(object1, object2)` detects no change.

## Examples more and more complexe

**At level 1**, directly comparing two `Set` having same elements but in a different order,  **Javers works as expected** : 

```java
@Test
public void compareSetsOfElementsInDifferentOrder() {
    // GIVEN Two Sets with elements in different order
    Set<String> streets1 = Set.of("Street A", "Street B");
    Set<String> streets2 = Set.of("Street B", "Street A");

    // WHEN compare with javers
    Diff diff = JAVERS.compare(streets1, streets2);

    // THEN we expect no change to be detected
    Assertions.assertFalse(diff.hasChanges());
}
```

**At level 2**, comparing two Object containing `Set`,  **Javers works as expected** : 

```java
@Test
public void compareObjectsContainingSetOfElementsInDifferentOrder() {
    // GIVEN Two objects containing Set with elements in different order
    City city1 = new City(Set.of("Street A", "Street B"));
    City city2 = new City(Set.of("Street B", "Street A"));

    // WHEN compare with javers
    Diff diff = JAVERS.compare(city1, city2);

    // THEN we expect no change to be detected
    Assertions.assertFalse(diff.hasChanges());
}
```

**At level 3**, comparing Set of ...... , **Javers works as expected** : 

```java
@Test
public void compareSetsOfObjectsContainingSetOfElementsInDifferentOrder() {
    // GIVEN Two sets of objects containing Set with elements in different order
    Set<City> cities1 = Set.of(new City(Set.of("Street A", "Street B")));
    Set<City> cities2 = Set.of(new City(Set.of("Street B", "Street A")));

    // WHEN compare with javers
    Diff diff = JAVERS.compare(cities1, cities2);

    // THEN we expect no change to be detected
    // Be careful : Methods City.equals() and City.hashCode() MUST be override
[...]

PR #1487, opened 2026-06-16
javers#1484 - Comparison of complex object with Set

Adrien-dev25:
Test case that highlights the problem described in the issue https://github.com/javers/javers/issues/1484

### DIFF
--- a/javers-core/src/main/java/org/javers/core/graph/SnapshotObjectHasher.java
+++ b/javers-core/src/main/java/org/javers/core/graph/SnapshotObjectHasher.java
@@ -1,11 +1,24 @@
 package org.javers.core.graph;
 
+import java.util.Comparator;
 import java.util.List;
 import java.util.stream.Collectors;
 
 import org.javers.common.string.ShaDigest;
 import org.javers.core.json.JsonConverter;
+import org.javers.core.metamodel.object.CdoSnapshotState;
+import org.javers.core.metamodel.type.CollectionType;
+import org.javers.core.metamodel.type.JaversProperty;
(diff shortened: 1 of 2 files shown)


bartoszwalacik:
@Adrien-dev25 , friendly reminder, your commits are not likned to your GH profile, so you won't be counted as a contributor. You can fix it in you GH profile

bartoszwalacik:
@Adrien-dev25 thanks for driving this, I proposed the less invasive solution , see https://github.com/javers/javers/pull/1493/changes

the case is we have to be sure we don't change by accident the hasing function for Value objects without Sets, which are out of scope of the that fix. The Hash is used as ID so we have to be extra carefull while changing it

Adrien-dev25:
> @Adrien-dev25 , friendly reminder, your commits are not likned to your GH profile, so you won't be counted as a contributor. You can fix it in you GH profile

Thanks I added my email address, it should be good now.

I'm afraid we've introduced an UnsupportedCastException for  `ListAsSetType` :
``` java
if (property.getType() instanceof SetType) {
                stateBuilder.withPropertyValue(property, toOrderedSet((Set<?>)propertyValue));
            }
            else if (property.getType() instanceof ListAsSetType) {
                stateBuilder.withPropertyValue(property, new TreeSet<>((List<?>) propertyValue));
            } 
            else {
                stateBuilder.withPropertyValue(property, propertyValue);
            }
```

I'll try to propose a fix

bartoszwalacik:
right, no test case for that

Adrien-dev25:
Fixed in this PR : https://github.com/javers/javers/pull/1494

PR #1489, opened 2026-06-27
Add property to control pro banner print

### LINKED ISSUE #1488: Disable promo banner
We would like disable the Javers Pro Promo banner

nstdio:
Closes: #1488


bartoszwalacik:
One issue with the current implementation: the default value of `printProBanner` is `false`, which means the banner is silently disabled for *all* existing users — not just those who opt out. That's a breaking change in behavior.

The field should default to `true` to preserve the existing behavior:

```java
private boolean printProBanner = true;
```

This way the banner keeps printing as before, and only users who explicitly call `.withPrintProBanner(false)` will suppress it — proper opt-out.

Also, could you wire this through `JaversCoreProperties` as well? Users of the Spring Boot starter would expect to be able to set `javers.printProBanner=false` in `application.properties` without touching Java config.

nstdio:
I think we just use previous version without banner. thanks.

Issue #1490, opened 2026-06-29
Javers detects changes when comparing nested Lists with `ListCompareAlgorithm.AS_SET`


Adrien-dev25:
## Expected Behavior
Since `ListCompareAlgorithm.AS_SET` is set, Javers should treat all lists, **including nested ones**, as unordered collections when comparing objects. 

At level 1, when I compare `List.of("Street A", "Street B")` and `List.of("Street B", "Street A")`, javers detect no changes which **is correct**.
But when I compare `List.of(List.of("Street A", "Street B"))` and `List.of(List.of("Street B", "Street A"))` **changes are detected**.

## Example
```java
private final static Javers JAVERS = JaversBuilder.javers()
                                                      .withListCompareAlgorithm(ListCompareAlgorithm.AS_SET)
                                                      .build();

@Test
 public void compareNestedListsWithDifferentOrderUsingAsSetAlgorithm() {
      // GIVEN deux nested Lists containing same elements but in a different ordre
      List<List<String>> streetsOriginalOrder = List.of(List.of("Street A", "Street B"));
      List<List<String>> streetsReversedOrder = List.of(List.of("Street B", "Street A"));

      // WHEN compare with javers setup with ListCompareAlgorithm.AS_SET
      Diff diff = JAVERS.compare(streetsOriginalOrder, streetsReversedOrder);

      // THEN we expect no change to be detected
      Assertions.assertFalse(diff.hasChanges());
 }
```

bartoszwalacik:
This is a known limitation of JaVers.

The key point is that JaVers is **statically typed**. It builds its type model from Java's reflection metadata, and reflection is performed **once for each class when that class is encountered for the first time in the object graph**. It is **not** performed for every property value of every object during comparison.

Because of that, when JaVers encounters a field declared as:

```java
List<Object> list;
```

the only type information available from the Java Reflection API is `Object`. At runtime, even if the list happens to contain instances of `A`, `B`, or `C`, JaVers doesn't inspect every element to infer a more specific type. It relies on the declared type of the field.

You can actually see this in the warning that JaVers logs:

```text
WARN  CorePropertyChangeAppender - Unknown item type in Set property:
Field list List<Object>, javersType: ListAsSetType,
declared in: ListWrapper.
Defaulting to Object, see CorePropertyChangeAppender.GENERIC_TYPE_NOT_PARAMETRIZED
```

That warning indicates that JaVers has no generic type information available and therefore defaults to `Object`.

[...]

PR #1491, opened 2026-06-29
javers#1490 - Comparison of nested List with ListCompareAlgorithm.AS_SET

Adrien-dev25:
Test case that highlights the problem described in the issue #1490


bartoszwalacik:
rejected, see explanation https://github.com/javers/javers/issues/1490

PR #1494, opened 2026-07-07
Fix ClassCastException when comparing nested List with ListCompareAlgorithm.AS_SET

Adrien-dev25:


bartoszwalacik:
thanks @Adrien-dev25

bartoszwalacik:
relesed in 7.11.6

PR #1496, opened 2026-07-14
#1488 add property to control pro banner print

### LINKED ISSUE #1488: Disable promo banner

tlopesPT:
Closes #1488.

Supersedes #1489, which was closed unmerged. Same approach as @nstdio's original, with both review comments from that PR applied:

1. `printProBanner` defaults to `true`, so existing behaviour is unchanged — the banner keeps printing for everyone who doesn't explicitly opt out.
2. The switch is wired through `JaversCoreProperties`, so it can be set from `application.yml` in the Spring Boot starter. `JaversSpringProperties` already extends `JaversCoreProperties`, so no Spring-side changes were needed.

We appreciate the continuous effort put into this library but would kindly request a way to disable the banner.


bartoszwalacik:
ok, no worries, I will merge

bartoszwalacik:
7.11.7

PR #1498, opened 2026-08-04
#1076 convert MissingProperty to null before CustomPropertyComparator

### LINKED ISSUE #1076: Exception during comparing String value - java.lang.ClassCastException: org.javers.core.metamodel.property.MissingProperty incompatible with java.lang.String
Using Javers I created custom comparator for String value so that null and empty String treat as same value. During manual tests in private project (I cant show you code of this project) I came accross such an exception:

> java.lang.ClassCastException: org.javers.core.metamodel.property.MissingProperty incompatible with java.lang.String
> at .StringComparator.compare(StringComparator.java:21)
> at org.javers.core.diff.custom.CustomToNativeAppenderAdapter.calculateChanges(CustomToNativeAppenderAdapter.java:33)
> at org.javers.core.diff.DiffFactory.appendChanges(DiffFactory.java:153)
> at org.javers.core.diff.DiffFactory.appendPropertyChanges(DiffFactory.java:143)
> at org.javers.core.diff.DiffFactory.createAndAppendChanges(DiffFactory.java:125)
> at org.javers.core.diff.DiffFactory.create(DiffFactory.java:69)
> at org.javers.core.diff.DiffFactory.compare(DiffFactory.java:54)
> at org.javers.core.JaversCore.compare(JaversCore.java:175)
> ...

This is what StringComparator look like:

```
class StringComparator implements CustomPropertyComparator<String, SetChange> {

    StringComparator() {
    }
    
    @Override
    public Optional<SetChange> compare(String originString, String modifizierteString,
            PropertyChangeMetadata propertyChangeMetadata, Property property) {
        if (equals(originString, modifizierteString)) {
            return Optional.empty();
        } else {
            final Set<Character> leftSet = getSetCharacterFromString(originString);
            final Set<Character> rightSet = getSetCharacterFromString(modifizierteString);

            return getSetChangeFromDiffrentSets(leftSet, rightSet, propertyChangeMetadata);
        }
    }

    @Override
    public boolean equals(String s1, String s2) {
        return StringUtils.equals(StringUtils.defaultString(s1), StringUtils.defaultString(s2));
    }

    @Override
    public String toString(String s) {
        return s;
    }
    
    private Set<Character> getSetCharacterFromString(String s) {
        return !isNull(s) ? s.chars().mapToObj(c -> (char) c).collect(Collectors.toSet()) : null;
    }
}
```
Can anyone tell me whether this is a javers bug or my code bug? I think I am not able to debug it, because immediately throws exception.

**Javers' Version**
5.14

kalayciburak:
## Summary
- Convert internal `MissingProperty` to `null` before invoking user `CustomPropertyComparator` implementations
- Prevents `ClassCastException` when a property exists on only one side of a polymorphic compare
- Aligns with `Atomic.unwrap()` treating `MissingProperty` as null
- Regression tests in `Case1076CustomPropertyComparatorMissingProperty`

Fixes #1076

## Test plan
- [x] `./gradlew :javers-core:test --tests org.javers.core.cases.Case1076CustomPropertyComparatorMissingProperty` (2/2)
- [x] Related custom comparator suites (CustomPropertyComparatorE2ETest, cases, example) green
- [x] `./gradlew :javers-core:test` — 1504 tests green (Java 17)

### DIFF
--- a/javers-core/src/main/java/org/javers/core/diff/changetype/Atomic.java
+++ b/javers-core/src/main/java/org/javers/core/diff/changetype/Atomic.java
@@ -42,7 +42,7 @@ public class Atomic implements Serializable {
      * original Value
      */
     public Object unwrap() {
-        return MissingProperty.INSTANCE == value ? null : value;
+        return MissingProperty.unwrap(value);
     }
 
     @Override
--- a/javers-core/src/main/java/org/javers/core/diff/custom/CustomToNativeAppenderAdapter.java
+++ b/javers-core/src/main/java/org/javers/core/diff/custom/CustomToNativeAppenderAdapter.java
@@ -3,6 +3,7 @@ package org.javers.core.diff.custom;
 import org.javers.core.diff.NodePair;
 import org.javers.core.diff.appenders.PropertyChangeAppender;
 import org.javers.core.diff.changetype.PropertyChange;
+import org.javers.core.metamodel.property.MissingProperty;
 import org.javers.core.metamodel.type.JaversProperty;
 import org.javers.core.metamodel.type.JaversType;
 
@@ -25,10 +26,22 @@ public class CustomToNativeAppenderAdapter<T, C extends PropertyChange> implemen
 
     @Override
     public C calculateChanges(NodePair pair, JaversProperty property) {
-        T leftValue = (T)pair.getLeftPropertyValue(property);
-        T rightValue = (T)pair.getRightPropertyValue(property);
+        T leftValue = sanitizeMissingProperty(pair.getLeftPropertyValue(property));
+        T rightValue = sanitizeMissingProperty(pair.getRightPropertyValue(property));
 
         return delegate.compare(leftValue, rightValue, pair.createPropertyChangeMetadata(property), property).orElse(null);
+    }
+
+    /**
+     * {@link MissingProperty} is an internal marker for "property present on only one side".
+     * It must not leak into user {@link CustomPropertyComparator} implementations (see #1076).
+     */
+    @SuppressWarnings("unchecked")
+    private T sanitizeMissingProperty(Object value) {
+        if (value == MissingProperty.INSTANCE) {
+            return null;
+        }
+        return (T) value;
     }
 
     @Override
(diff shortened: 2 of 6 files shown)


bartoszwalacik:
Hi @kalayciburak 

Thanks for tracking down the root cause here and adding a repro — this is a clean fix, and the `sanitizeMissingProperty` approach makes sense, especially the alignment with how `Atomic.unwrap()` already treats `MissingProperty`.

A couple of small things before merging:

1. **`CustomValueComparatorNullSafe.toString()`**
   This hunk isn't covered by the new test (`Case1076CustomPropertyComparatorMissingProperty` only exercises the `CustomPropertyComparator` path via `CustomToNativeAppenderAdapter` / `CustomType`). Could you either:
   - add a quick case for it, or
   - split it into a follow-up PR?

   Just want every changed line to have a test backing it.

2. **Duplicated helper**
   `sanitizeMissingProperty()` appears verbatim in both `CustomToNativeAppenderAdapter` and `CustomType`. Would you mind pulling it into a single static method — maybe right on `MissingProperty` itself, e.g.:
```java
   MissingProperty.unwrap(Object value)
```
   Keeps the logic in one place if we ever need to touch it again.

Nothing structural — the core fix and test for #1076 look right to me. Happy to take another look once these are in 🙂

kalayciburak:
Thanks for the review @bartoszwalacik — both points addressed in 8c35fe00:

1. **`CustomValueComparatorNullSafe.toString()` coverage**  
   Added a focused case in `Case1076CustomPropertyComparatorMissingProperty` that hits `valueToString(MissingProperty.INSTANCE)` via `registerValue` + a strict `CustomValueComparator` (would CCE if `MissingProperty` leaked).

2. **Shared helper**  
   Extracted `MissingProperty.unwrap(Object)` and routed:
   - `CustomToNativeAppenderAdapter`
   - `CustomType.CustomPropertyComparatorNullSafe#compare`
   - `CustomValueComparatorNullSafe#toString`
   - `Atomic#unwrap` (same semantics as before)

Local check (Java 17):
```
./gradlew :javers-core:test \
  --tests org.javers.core.cases.Case1076CustomPropertyComparatorMissingProperty \
  --tests org.javers.core.cases.CustomValueComparatorMissingPropertyCase \
  --tests org.javers.core.CustomPropertyComparatorE2ETest
```
→ 28 tests green.

Happy to tweak further if anything else stands out.

bartoszwalacik:
Thanks @kalayciburak , will merge and release

bartoszwalacik:
released in 7.11.8

